Issue #183, opened 2013-11-29
--follow does not show redirects


vartec:
I've expected `http --follow` to act in similar fashion as `curl -i -L`, however difference is that curl does show headers of all intermediate steps, httpie does not.

Expected result (standard post-redirect-get form):
```
HTTP/1.1 100 Continue

HTTP/1.1 303 See Other
Content-Type: text/html; charset=UTF-8
Date: Fri, 29 Nov 2013 11:50:59 GMT
Location: http://other.test.server/
Server: nginx/1.1.19
Content-Length: 0
Connection: keep-alive

HTTP/1.1 200 OK
Date: Fri, 29 Nov 2013 11:51:00 GMT
Server: Apache/2.2.14 (Ubuntu)
Vary: Accept-Encoding,Cookie
Pragma: no-cache
Cache-Control: private, max-age=0, must-revalidate
Transfer-Encoding: chunked
Content-Type: application/json

{"payload": "something"}
```
Actual result:
```
HTTP/1.1 200 OK
Cache-Control: private, max-age=0, must-revalidate
Content-Encoding: gzip
Content-Length: 414
Content-Type: application/json
Date: Fri, 29 Nov 2013 11:55:34 GMT
Pragma: no-cache
Server: Apache/2.2.14 (Ubuntu)
Vary: Accept-Encoding,Cookie

{"payload": "something"}
```

ralphschindler:
:+1:

rockymadden:
Looking for this as well. Currently using a custom bash function to accommodate.

ian-rose:
This would be _very_ useful functionality to have.

tylercrompton:
I'll redo this if @jakubroztocil would review the changes. But after a year, there were too many conflicts and it didn't seem like it would ever get reviewed, so I just deleted what I did.

dstarmanhr:
this would be a very useful feature to have.

jlazic:
+1 for this feature.

jonoden:
+1

jbenoit2011:
+1 for that feature. Thanks.

sgoblin:
Can you give a brief summary of how you implemented this, @tylercrompton ?

jcdevil:
+1 for that feature. Thanks ;)

victorgama:
+1 for that.

Wirone:
[...]

PR #186, opened 2013-12-16
Solves Issue #183 (--follow does not show redirects)

tylercrompton:
I implemented in a way different than described (cURL's implementation). I implemented it so that it respected the output formatting arguments. Perhaps we can add a `--last` argument to only output the last response? That wouldn't too much work to implement.

We still need to write one or more tests for this, though.

This is an extension of Issue #184.

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -57,6 +57,8 @@ def get_responses(args, config_dir):
             read_only=bool(args.session_read_only),
         )
 
+    # Yielding the response before retrieving the next response (if any)
+    # causes an error.
     redirect = None
     redirects = ()
     if args.follow:


Remzi1993:
Hi @tylercrompton 
I think this isn't implemented. Are you still working on this?

Kind regards,
Remzi

tylercrompton:
@Remzi1993, no, I'm sorry. I admittedly never implemented any tests [as he requested](https://github.com/jakubroztocil/httpie/pull/184#issuecomment-30609784), but my questions in #184 went unanswered, so I didn't want to make the decisions myself since changes according to such decisions could have been rejected potentially. Therefore, my existing changes remained in purgatory and thus never got merged.

Because this was several years ago, I believe that aside from the tests, the only work that was left to be done pertained to the questions that I asked in #184. Specifically, the important unanswered questions were as follows:

> Perhaps we can add a --last argument to only output the last response?

> I'm starting to think that by default it should only print the last response.

These are the decisions that @jakubroztocil never made. Regarding my remarks about the `-v`/`--verbose` flags, I now believe that always printing all responses would be the logical decision.

I've now restored [the branch](https://github.com/tylercrompton/httpie/commits/follow_headers) so that you can take a look at it if you want, but as [I said over four years ago](https://github.com/jakubroztocil/httpie/issues/183#issuecomment-117714677), the merge conflicts were so big that the work would have had to be redone from scratch. And that surely hasn't changed since then. Because my questions still remain unanswered, I unfortunately still see no reason to revisit this.

At the very least, I suppose that you could attempt to use the program as-is from the tip of my branch, but obviously, it will be missing nearly six years of updates. Looking at "setup.py", HTTPie looks like it was compatible with Python 2.7 and 3.2 at the time. Use the "new" `--follow` option when executing HTTPie. I understand this probably isn't the response for which you were hoping but that's all that I've got. Sorry.

Remzi1993:
Hi @tylercrompton 
Thank you for your detailed reply!
I hope @jakubroztocil could do something about this. I think the default should be to show every redirecting and the headers of that redirection just like to browsers to by default instead of showing only the end result.

PR #229, opened 2014-06-20
Read password from environment variable

gmanika:
If the password is not provided via the command line, try to read it from the HTTPIE_PASSWORD environment variable.

This is meant as a simple way to avoid leaking the password via the process title in scripted runs.


sirianni:
Is this functionality available in some other form?

PR #236, opened 2014-06-25
+Dockerfile

schvin:
Add a simple dockerfile for arbitrary usage.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM ubuntu:trusty
+MAINTAINER George Lewis <schvin@schvin.net>
+RUN apt-get update --fix-missing
+
+RUN apt-get install -y python-pip
+RUN pip install --upgrade httpie
+
+RUN groupadd httpie
+RUN useradd httpie -g httpie -d /home/httpie
+RUN mkdir /home/httpie
+RUN chown -R httpie:httpie /home/httpie
+
+ENV HOME /home/httpie
+USER httpie
+WORKDIR /home/httpie
+
+CMD ["--help"]
+ENTRYPOINT ["/usr/local/bin/http"]


flaccid:
+1 to this.

jirutka:
-1 for this, it’s extreme overkill to run whole bloated Ubuntu in Docker just to use simple Python utility.

mfussenegger:
If this gets included I'd at least recommend to optimize the Dockerfile a bit. 

The RUN commands to add the httpie user/group all cause extra layers which are very unlikely to be shared. See [Dockerfile best practices](https://docs.docker.com/engine/articles/dockerfile_best-practices/).
And using Ubuntu as base is probably not the best option either. There are [dedicated python images](https://hub.docker.com/_/python/). Especially python:alpine might be interesting as it would result in a much smaller image.

schvin:
definitely agree - those images and such were not available 2 years ago when this PR was submitted, so it could definitely use some improvements ;)

given the time that has passed i assumed there was no interest. happy to clean it up and trim it down if others want it - i use it all the time and find it super useful!

jkbrzt:
@schvin thanks but I don't think including it inside the repo adds a lot of value

flaccid:
Work in progress: https://github.com/flaccid/docker-httpie. Happy to take contribution and will maintain it.

hoatle:
I added this: https://github.com/teracyhq/docker-files/tree/master/httpie-jwt-auth that you can use the built Docker image right away

https://hub.docker.com/r/teracy/httpie-jwt-auth/tags/

I'm using this Docker image and it works great. I'll find a way to build official Docker image for this project with docs (as we did for https://github.com/teracyhq/docker-files/tree/master/httpie-jwt-auth)

Let me create an issue for this first.

hoatle:
issue created: https://github.com/jakubroztocil/httpie/issues/571

PR #246, opened 2014-07-30
Implemented --max-redirects option from issue 188

JamesDavidCarr:
Added argument to argparse, changed client so that it uses a new
requests Session() with the number of redirects and a single test to
show the setting works.


JamesDavidCarr:
@jkbrzt I just found this commit from a while ago, has this issue been resolved?

PR #273, opened 2014-11-12
Added test for overriding Content-Length.

asnelzin:
Related to https://github.com/jakubroztocil/httpie/issues/269

Just a test to be sure then requests will be updated.


jakubroztocil:
I think this is not relevant anymore. But thanks!

Issue #209, opened 2014-04-04
support http over unix domain sockets


monsanto:
For security reasons, some of my HTTP services need to run on a unix domain socket so I can protect them using filesystem permissions. It would be useful if I could use httpie for testing these services.

docker-py uses the requests library and has support for this--see https://github.com/dotcloud/docker-py/blob/master/docker/unixconn/unixconn.py for the (small) implementation, and https://github.com/dotcloud/docker-py/blob/master/docker/client.py#L90 for how they extend requests using `self.mount`.

jakubroztocil:
Implemented as as plugin (supported by HTTPie v0.9.1+): [httpie-unixsocket](https://github.com/msabramo/httpie-unixsocket)

PR #276, opened 2014-11-25
Add support for unix domain sockets

msabramo:
To issue HTTP requests to a UNIX domain socket:

* Use new `"http+unix://"` scheme in URL.
* Use urlencoded socket path as the hostname part of the URL.
  e.g.: for a socket at `/tmp/profilesvc.sock`, you could do:

      http http+unix://%2Ftmp%2Fprofilesvc.sock/status/pid

This uses https://pypi.python.org/pypi/requests-unixsocket

Fixes: GH-209 (https://github.com/jakubroztocil/httpie/issues/209)

#### Screenshot

screen shot 2014-11-24 at 6 35 24 pm

Cc: @shin-, @jakubroztocil, @monsanto, @np, @nuxlli, @matrixise, @remmelt

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -298,6 +298,16 @@ command:
 
     GET /?search=HTTPie&tbm=isch HTTP/1.1
 
+To issue HTTP requests to a UNIX domain socket (added in `0.9.0-dev`_):
+
+* Use new ``http+unix://`` scheme in URL.
+* Use url-encoded socket path as the hostname part of the URL.
+  e.g.: for a socket at ``/tmp/profilesvc.sock``, you could do:
+
+.. code-block:: bash
+
+    $ http http+unix://%2Ftmp%2Fprofilesvc.sock/status/pid
+
 
 =============
 Request Items
@@ -1329,6 +1339,8 @@ Changelog
     * Fixed handling of empty passwords in URL credentials.
     * Fixed multiple file uploads with the same form field name.
     * Fixed ``--output=/dev/null`` on Linux.
+    * Improved terminal color depth detection via ``curses``.
+    * Add support for unix domain sockets with ``http+unix://`` scheme.
     * Miscellaneous bugfixes.
 * `0.8.0`_ (2014-01-25)
     * Added ``field=@file.txt`` and ``field:=@file.json`` for embedding
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -3,6 +3,7 @@ import sys
 from pprint import pformat
 
 import requests
+import requests_unixsocket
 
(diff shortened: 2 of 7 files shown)


jakubroztocil:
@msabramo what do you think about adding this as a plugin (https://github.com/jakubroztocil/httpie/pull/276#issuecomment-64484844)? There is also another plugin candidate that is basically a requests adapter (#298).

msabramo:
I'm open to creating an HTTPie plugin. I need some guidance though.

From a glance at the code, it looks like there are 3 types of plugins:

1. auth
2. converter
3. formattter

If I am not mistaken, these all get called **after** the request has already completed. In that case that would be too late to try to open a unix socket.

Does there need to be a new type of plugin perhaps -- a connection plugin?

msabramo:
Actually it's looking like I can maybe make it an auth plugin and then register my requests adapter at the top level of the module so that it executes at import time. But this feels a little hacky so maybe a connection plugin is still a good idea.

msabramo:
@jakubroztocil: Please see https://github.com/jakubroztocil/httpie/pull/299

With that, I can get an auth plugin to work.

msabramo:
And here's the unix socket "auth plug" I hacked together:

https://github.com/msabramo/httpie-unixsocket

I guess it would be nice if it were a "connection plugin" rather than an "auth plugin", but we'd have to add support to HTTPie for that.

jakubroztocil:
@msabramo

> From a glance at the code, it looks like there are 3 types of plugins:
> 1. auth (the only kind that is documented in README.rst by the way)
> 2. converter (undocumented)
> 3. formattter (undocumented)

That is correct. The plugin API is still work in progress and will probably be completely reworked by v1.0.0, that's why it's lacking docs. But for now, I will add a new plugin type, an `AdapterPlugin` ([`python-requests`: transport adapters](http://docs.python-requests.org/en/latest/user/advanced/#transport-adapters)). HTTPie will load these plugins on startup and mount them (just like you did in the first version of this pull request).

jakubroztocil:
@msabramo Let me please know if it's okay. I'll then release v0.9.1 and you can publish the plugin on PyPi.

jakubroztocil:
@msabramo HTTPie v0.9.1 with transport adapter plugin support is out. So the plugin can be released on PyPi as well (after https://github.com/msabramo/httpie-unixsocket/pull/3 has been merged).

jakubroztocil:
Implemented as as plugin supported in HTTPie v0.9.1+:  https://github.com/msabramo/httpie-unixsocket

PR #284, opened 2014-12-04
Implement basic msgpack support as a converter

rasky:
One year ago, I submitted msgpack support before the plugin system (see #176). I'm now resubmitting this as a formatter plugin.


jakubroztocil:
@rasky This would be better to have as a plug-in (`httpie-msgpack`) and put it on pypi with `msgpack` as a requirement.

rasky:
Done: https://github.com/rasky/httpie-msgpack

Issue #243, opened 2014-07-19
Response time


whodidthis:
Could be handy to see a requests response time in verbose or with some other flag.

guilleferrer:
+1

PavloVasylchenko:
+1

matrixise:
+1

bgalek:
+1

RyanDhuse:
+1

exegete46:
+1

ravshansbox:
+1

asychev:
+1

gblmarquez:
👍

nowtvicebergteam:
👍

heyalexej:
+1

oleggromov:
I'd like to have response time too! 👍

aum12:
+1

qoomon:
+1

postmeridiem:
+1

ahbruinsma:
+1

Evlos:
+1

kusha:
+1

davegallant:
+2

tmhpfnr:
+1

sahas-:
+1

andreafalzetti:
+1

bytedaring:
+1

bytedaring:
👍

lcd1232:
👍

stellarpower:
For anyone looking for a quick hack that may get the job done, the bonus is that I learned today about the `time` command in Fish (and possibly your other favourite shell, I believe the bash has it)

ramilmsh:
> For anyone looking for a quick hack that may get the job done, the bonus is that I learned today about the `time` command in Fish (and possibly your other favourite shell, I believe the bash has it)

JFYI, this will time the command, not request, meaning parsing the arguments, making the request, compression, parsing response and so on will be included. It's fine, if your request time much larger (orders of magnitude) than any local operations that need to be performed, but something you need to keep an eye out for

jakubroztocil:
🎉 This feature is now publicly available in [HTTPie for Terminal 3.0](https://httpie.io/blog/httpie-3.0.0): 

👉 https://httpie.io/docs/cli/response-meta

PR #286, opened 2014-12-16
Addresses Issue #243

dlynch158:
This is my attempt at addressing issue #243. It adds a 't' option to the -p (print) command so that the elapsed time can be displayed at the end of the output.

It works as-is with one problem: the body gets printed by default (even with JUST "-p t"). I wasn't able to track down why that was happening.


msabramo:
See https://github.com/dlynch158/httpie/pull/1

PR #299, opened 2015-02-04
Loosen check for "http" scheme in URL

msabramo:
```diff
-        if not (self.args.url.startswith((HTTP, HTTPS))):
+        if not re.match(r'^http.*://', self.args.url):
```

This allows us to accommodate new connection schemes such as the "http+unix" scheme being discussed in https://github.com/jakubroztocil/httpie/pull/276


jakubroztocil:
@msabramo I've added a bit more robust scheme handling (it will allow non-http custom schemes as well).

PR #300, opened 2015-02-04
Print info about request on error

msabramo:
This can help in diagnosing certain issues. For example, if I were
trying to use a "http+unix" URL but I don't have #299, then I'll get the
following:

    [marca@marca-mac2 httpie]$ http http+unix://%2Ftmp%2Fprofilesvc.sock/status/pid

    http: error: ConnectionError: ('Connection aborted.', gaierror(8, 'nodename nor servname provided, or not known'))
    while doing GET request to URL: http://http+unix//%2Ftmp%2Fprofilesvc.sock/status/pid

Having the URL in the error message is super useful here so that I know an
extra `http://` is getting prepended and it's not doing what I expected.

PR #302, opened 2015-02-09
Print info about request on error 2

msabramo:
Test to see if this decreases test coverage because I can't figure out why  coveralls is saying that #300 has decreased coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,4 @@ python:
 script:
   - make
 after_success:
-  - pip install python-coveralls
-  - coveralls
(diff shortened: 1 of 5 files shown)


msabramo:
@coveralls: I cannot reproduce a coverage drop here. I thought I added coverage actually. Any ideas?

jakubroztocil:
@msabramo it's probably because it gets reported after each build, and every time it can be a different Python version that finishes last. There some version-specific code paths (eg. in compact.py).

msabramo:
Oops probably should've used 2.6 since that had the highest coverage

PR #303, opened 2015-02-10
.travis.yml: Only do coveralls on one Python version

msabramo:
Testing theory that it has to do with different python version subjobs completing in different orders and the last one wins.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@
 os:
   - linux
   - osx
+env:
+  global:
+    - NEWEST_PYTHON=3.4
 language: python
 python:
   - 2.6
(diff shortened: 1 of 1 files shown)


jakubroztocil:
Great, this will rid us of the confusing coverage drops/increases. 

I'd go for the highest version though (currently `3.4`). And instead of hard-coding the version numbe in the condition, I'd prefer something more explicit, such as:

```yml
env:
  global:
    - NEWEST_PYTHON=3.4
```
…
```yml
 after_success
  - if [[ $TRAVIS_PYTHON_VERSION == $NEWEST_PYTHON ]]; then pip install python-coveralls && coveralls; fi
```

Even better would be to pick the highest version directly from the `python:` list in the `.travis.yml`, but I'm not sure if that is possible.

(Btw, I see that the `test_session_unicode`-related commit has its own pull request [#305], but it's also included in this one as well as in #304.)

msabramo:
Updated.

jakubroztocil:
Thanks :+1:

PR #304, opened 2015-02-10
compat.py: Add pragma no covers

msabramo:
Cuz this is a lot of version-specific stuff and it can be confusing to have different coverage per version, especially with coveralls.


jakubroztocil:
:+1: 

Would you just please remove 5b3d749 from this pull request?

msabramo:
> Would you just please remove 5b3d749 from this pull request?

Done.

jakubroztocil:
:+1:

Issue #282, opened 2014-12-01
py34 test failure: KeyError: 'Authorization' error in TestSession.test_session_unicode


msabramo:
py34 test failure: `KeyError: 'Authorization'` error in `TestSession.test_session_unicode`

I can reproduce the `test_session_unicode` failure consistently by explicitly passing a `--hashseed` to tox:

```
❯ tox -e py34 --hashseed=1811760512 -- tests/test_sessions.py -k test_session_unicode
GLOB sdist-make: /Users/marca/dev/git-repos/httpie/setup.py
py34 inst-nodeps: /Users/marca/dev/git-repos/httpie/.tox/dist/httpie-0.9.0-dev.zip
py34 runtests: PYTHONHASHSEED='1811760512'
py34 runtests: commands[0] | py.test --verbose --doctest-modules --basetemp=/Users/marca/dev/git-repos/httpie/.tox/py34/tmp tests/test_sessions.py -k test_session_unicode
============================================================================= test session starts ==============================================================================
platform darwin -- Python 3.4.0 -- py-1.4.26 -- pytest-2.6.4 -- /Users/marca/dev/git-repos/httpie/.tox/py34/bin/python3.4
plugins: httpbin
collected 6 items

tests/test_sessions.py::TestSession::test_session_unicode FAILED

=================================================================================== FAILURES ===================================================================================
_______________________________________________________________________ TestSession.test_session_unicode _______________________________________________________________________
Traceback (most recent call last):
  File "/Users/marca/dev/git-repos/httpie/tests/test_sessions.py", line 151, in test_session_unicode
    assert (r2.json['headers']['Authorization']
KeyError: 'Authorization'
----------------------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------------------
127.0.0.1 - - [29/Nov/2014 11:40:28] "GET /get HTTP/1.1" 200 301
127.0.0.1 - - [29/Nov/2014 11:40:28] "GET /get HTTP/1.1" 200 301
================================================================ 5 tests deselected by '-ktest_session_unicode' ================================================================
============================================================== 1 failed, 5 deselected, 1 warnings in 0.67 seconds ==============================================================
ERROR: InvocationError: '/Users/marca/dev/git-repos/httpie/.tox/py34/bin/py.test --verbose --doctest-modules --basetemp=/Users/marca/dev/git-repos/httpie/.tox/py34/tmp tests/test_sessions.py -k test_session_unicode'
[...]

PR #305, opened 2015-02-10
Mark test_session_unicode as xfail

msabramo:
There are known problems with unicode in headers

### DIFF
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -1,6 +1,9 @@
 # coding=utf-8
 import os
 import shutil
+import sys
+
+import pytest
 
 from httpie.plugins.builtin import HTTPBasicAuth
 from utils import TestEnvironment, mk_config_dir, http, HTTP_OK, \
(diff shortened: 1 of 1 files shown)


msabramo:
Related: https://github.com/jakubroztocil/httpie/issues/282

jakubroztocil:
Looking at the documentation for [`pytest.mark.xfail`](http://pytest.org/latest/skipping.html#mark-a-test-function-as-expected-to-fail), it appears to me that when this decorator is used, then the test is expected to fail *every time*:

> This test will be run but no traceback will be reported when it fails. Instead terminal reporting will list it in the “expected to fail” or **“unexpectedly passing” sections**.

The problem we're having though is that it only fails *sometimes*, and when it does, we want to ignore the failure. But when it passes, it shouldn't be treated as an unexpected event.
 
I'd say marking it with `pytest.mark.skip` for Py3k (with a message linking to #282) might be the best option until the root cause is fixed.

jakubroztocil:
Also, I see `pypy` and `pypy3` have been removed — why?

msabramo:
Updated to use `pytest.mark.skipif` and I put `pypy` and `pypy3` back since that is a separate issue (you're right about why I removed them).

jakubroztocil:
Great, thanks!

PR #319, opened 2015-03-08
Update license with up-to-date year

fay-jai:
Only made 1 change:
1) Update LICENSE to include the latest year in the form of a date-range.

No other changes were made.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright © 2012 Jakub Roztocil <jakub@roztocil.co>
+Copyright © 2012-2015 Jakub Roztocil <jakub@roztocil.co>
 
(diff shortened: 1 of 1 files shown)


jakubroztocil:
Thanks @fay-jai. Does a date-range have any benefits over a single date? AFAIK when only one year is included, then it's implied that it's until `$NOW`.

fay-jai:
To be honest, I'm basing this off of what I've seen on other large scale projects. It seems like the standard protocol which is why I suggested the change as well!

leesei:
A date range can indicate when the content was first published and when it was last modified.
That has to do with counting years for copyright expiration.

jakubroztocil:
Ok, thanks :+1:

Issue #326, opened 2015-03-23
Bash tab completion


Mahoney:
Feature request: it would be nice if someone could pull together a bash extension to give tab completion when using httpie.

jakubroztocil:
Initial Bash auto completion has been added in #330

PR #330, opened 2015-03-25
Bash auto completion

mihirvj:
See #326 

Adds bash completion to http command line interface.

Installing the script:
You can copy it to /etc/bash_completion.d/ (or something else on your
machine) and source it using following command

	$ source /etc/profile

Now whenever you encounter a "-*" on your CLI, it presents you with the
options specified.

Couple of things that are still under work:
1) Adding this bash script to setup, so that user won't need manual
installation
2) Adding more options for HTTP (GET, PUT and so on) and other
options

### DIFF
--- a/httpie-completion.bash
+++ b/httpie-completion.bash
@@ -1,0 +1,22 @@
+#!/bin/env bash
+
+_http_complete() {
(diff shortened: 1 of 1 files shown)


jakubroztocil:
This is a good start, thanks! :+1:

Issue #331, opened 2015-03-25
Bash completion improvements


jakubroztocil:
- [ ] Add support for HTTP methods
- [ ] Add support for URLs: perhaps complete known hosts like `ping` or `ssh` and known URLs from various sources
- [ ] Add support for choice values (e.g. `-s <tab>`, `--style=<tab>`, `--auth-type<tab>`, etc.)
- [ ] Update the [Homebrew formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/httpie.rb) to install the completion support

// #330 #326

mihirvj:
I'll look into this over next few days

jakubroztocil:
This looks quite interesting: https://github.com/kislyuk/argcomplete

So does [the way Django does it](https://github.com/django/django/blob/4a0aeac1b5cfb7b6229a01119a596afb38d8a2a0/django/core/management/__init__.py#L191-L263) for `manage.py`.

mihirvj:
Yes. I was looking into argcomplete previously. Couple of issues I found were that requirement of a new package argcomplete. Also, it requires additional setup like activation of global python autocomplete. Wasn't entirely sure whether to go that way. Django way looks more interesting. 

I understand that writing a different bash script for autocompletion will add an overhead in case options change, so better to include it in python.

I'll see which one to dig deep into.

PR #332, opened 2015-03-27
Added support for HTTP methods

mihirvj:
See #331

I think before we finalize the pythonic method for this, these changes can be utilized.


sigmavirus24:
Can you change the commit message (and title) to read:

> "Added completion support for HTTP methods"

The message is *very* misleading otherwise.

jakubroztocil:
@mihirvj I'd also add `PATCH`, `OPTIONS`, and I believe there are other commonly used methods as well

jakubroztocil:
Now I realize this will only work without any `--options` (e.g. `$ http -v P<tab>` wont complete). 

The more robust approaches mentioned in #331 are probably a better way to go.

PR #333, opened 2015-03-27
Bash completion with argcomplete

mihirvj:
See #331 #332 
So, here is what argcomplete looks like. It is relatively easy but it
requires an additional "argcomplete" package available on pip.

Also, have you thought of adding an -m and --method option for HTTP
methods? Any specific reason it was not a part of option and just a
sub command like syntax?

### DIFF
--- a/httpie-completion.bash
+++ b/httpie-completion.bash
@@ -1,22 +1,3 @@
 #!/bin/env bash
 
-_http_complete() {
-	local cur_word=${COMP_WORDS[COMP_CWORD]}
-	local prev_word=${COMP_WORDS[COMP_CWORD - 1]}
-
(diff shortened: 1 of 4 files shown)


mihirvj:
I am not sure why the build is failing. The tests work fine on my machine. Can anybody tell me the issue? I am sorry but I am new to this project so kindly bear me. Thanks!

mihirvj:
@jakubroztocil Can you help me? I have tried many things to prevent build from failing but I am really not sure of the issue. I think adding the package argcomplete is making build to fail but I am not sure how to prevent it from doing so.
The issue is harder to debug for me because all the tests (even if I uninstall argcomplete) are working correctly on my machine.
If you have seen such issue then kindly help me understand it.
Thanks

PR #337, opened 2015-04-11
Only serialize json if data is a dict instance

### LINKED ISSUE #336: Using --json with redirected input raises TypeError
```
$ echo '{"foo": "bar"}' | http --debug --json httpbin.org/post
HTTPie 0.9.2
HTTPie data: /home/joao/.httpie
Requests 2.6.0
Pygments 2.0.2
Python 3.4.3 (default, Mar 25 2015, 17:13:50) 
[GCC 4.9.2 20150304 (prerelease)] linux
  File "/home/joao/.local/bin/http", line 11, in <module>
    sys.exit(main())
  File "/home/joao/.local/lib/python3.4/site-packages/httpie/core.py", line 112, in main
    response = get_response(args, config_dir=env.config.directory)
  File "/home/joao/.local/lib/python3.4/site-packages/httpie/client.py", line 38, in get_response
    kwargs = get_requests_kwargs(args)
  File "/home/joao/.local/lib/python3.4/site-packages/httpie/client.py", line 97, in get_requests_kwargs
    data = json.dumps(data)
  File "/usr/lib/python3.4/json/__init__.py", line 230, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.4/json/encoder.py", line 192, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.4/json/encoder.py", line 250, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.4/json/encoder.py", line 173, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: b'{"foo": "bar"}\n' is not JSON serializable
```

Although the `--json` flag is not necessary here, I think this should be valid and not raise `TypeError`.

The problem seams to be [with this condition](https://github.com/jakubroztocil/httpie/blob/master/httpie/client.py#L95) as it tries to serialize the data even if  data isn't a `dict` like object when `args.json = True`

joaodelgado:
Only serialize json data if it's  a `dict`.

Fixes #336

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -92,7 +92,7 @@ def get_requests_kwargs(args, base_headers=None):
     # Serialize JSON data, if needed.
     data = args.data
     auto_json = data and not args.form
-    if args.json or auto_json and isinstance(data, dict):
+    if (args.json or auto_json) and isinstance(data, dict):
         if data:
             data = json.dumps(data)
         else:


jakubroztocil:
Good catch. Obrigado João! :+1:

PR #368, opened 2015-07-22
Add support for overriding mime type

jcrumb:
Added a mime override argument to the PrettyStream class to allow forcing pretty output when a server does not set the correct Content-Type. 

I'd like some feedback on what I have before going any further. It works, but is there a better way to integrate this?

### DIFF
--- a/httpie/cli.py
+++ b/httpie/cli.py
@@ -217,6 +217,16 @@ output_processing.add_argument(
     )
 )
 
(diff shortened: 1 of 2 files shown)


jkbrzt:
Now with the `--json, -j` option HTTPie tries to detect JSON responses when the `Content-Type` is unknown or `text/plain`.

Issue #342, opened 2015-04-22
Accept input from a pipe for a multipart form


0x6e6562:
I'm able to upload a file using the following form syntax:

```
$ http -f POST 127.0.0.1/upload file@some_file.txt
```

but I was wondering if there is a way to pipe stdout to the form, for example:

```
$ date | http -f POST 127.0.0.1/upload file@-
```

This results in the following error:

```
http: error: "file@-": [Errno 2] No such file or directory: u'-'
```

I can achieve the same thing with curl:

```
$ date | curl -XPOST 127.0.0.1/upload -H 'Content-Type:multipart/form-data' --form file=@-
```

So I was wondering whether I was doing something wrong with httpie.

rocketraman:
Another way to achieve this in `curl` without requiring a filename is to specify the `multipart/form-data` Content-Type header with a `-F` form field:

    curl -X POST 127.0.0.1/upload -H 'Content-Type: multipart/form-data' -F 'metadata={"description": "foo"}'

curl creates a part that looks like this:

```
--------------------------69798281f9911ef1
Content-Disposition: form-data; name="metadata"

{"description": "foo"}
```

The closest I can get with httpie is to use process substitution for the input:

    http --form POST 127.0.0.1:8080/upload metadata@<(echo -n '{"description": "foo"}')

which creates:

```
--a7311cce943f40c4837806985cff8404                                                                                                                                                        
Content-Disposition: form-data; name="metadata"; filename="63"                                                                                                                            
                                                                                                                                                                                          
{"description": "foo"}
```

Which is close, but note the filename added to the `Content-Disposition` from the process substitution file descriptor.

PR #370, opened 2015-07-27
Accept input from a pipe for a multipart form

keita1314:
Hey,jkbrzt,I am a big fan of httpie, so I try to contribute to your project.This commit is for #342.
We can use httpie to upload a anonymous file through input from a pipe like curl.
**curl**
 >$ date | curl -XPOST http://example.org/upload -H 'Content-Type:multipart/form-data' --form file=@-

**httpie** 
>$ date | http -f --debug --traceback POST http://example.org/upload file@-

### DIFF
--- a/httpie/input.py
+++ b/httpie/input.py
@@ -328,6 +328,22 @@ class Parser(ArgumentParser):
             self.args.files = items.files
             self.args.params = items.params
 
+        if self.args.files and '-' in self.args.files.values():
+            # input from a pipline for a multipartform
+            # ` cmd | http -f POST url file@-`
+            file_fields = list(self.args.files.keys())
+            if file_fields == [''] or self.args.files.values() != ['-']:
+                self.error(
+                    'Invalid file fileds(you just need one f@-'')'
+                )
+            if self.args.ignore_stdin or self.env.stdin_isatty:
+                self.error(
+                    'Invalid stdin(you need stdin when http -f url f@-)'
+                )
+            self.args.files.clear()
+            value = ('-', BytesIO(self.env.stdin.read()))
+            self.args.files.__setitem__(file_fields[0], value)
+
         if self.args.files and not self.args.form:
             # `http url @/path/to/file`
             file_fields = list(self.args.files.keys())
@@ -630,12 +646,15 @@ def parse_items(items,
         elif item.sep == SEP_QUERY:
             target = params
         elif item.sep == SEP_FILES:
-            try:
-                with open(os.path.expanduser(value), 'rb') as f:
-                    value = (os.path.basename(value),
-                             BytesIO(f.read()))
-            except IOError as e:
-                raise ParseError('"%s": %s' % (item.orig, e))
+            if(value == '-'):
+                pass
+            else:
+                try:
+                    with open(os.path.expanduser(value), 'rb') as f:
+                        value = (os.path.basename(value),
+                                BytesIO(f.read()))
+                except IOError as e:
+                    raise ParseError('"%s": %s' % (item.orig, e))
             target = files
 
         elif item.sep in SEP_GROUP_DATA_ITEMS:


httpie/input.py
```diff
@@ -328,6 +328,22 @@ def _parse_items(self):
...
+                self.error(
+                    'Invalid file fileds(you just need one f@-'')'
+                )
+            if self.args.ignore_stdin or self.env.stdin_isatty:
+                self.error(
+                    'Invalid stdin(you need stdin when http -f url f@-)'
+                )
+            self.args.files.clear()
+            value = ('-', BytesIO(self.env.stdin.read()))
+            self.args.files.__setitem__(file_fields[0], value)
```
sigmavirus24: Why is this calling `__setitem__`?

httpie/input.py
```diff
@@ -328,6 +328,22 @@ def _parse_items(self):
             self.args.files = items.files
             self.args.params = items.params
 
+        if self.args.files and '-' in self.args.files.values():
```
sigmavirus24: Does `cURL` allow for you to do something like:

```
date | curl ... file@- file@- file@-
```

If so, does this allow HTTPie to work the same way?

httpie/input.py
```diff
@@ -630,12 +646,15 @@ def parse_items(items,
         elif item.sep == SEP_QUERY:
             target = params
         elif item.sep == SEP_FILES:
-            try:
-                with open(os.path.expanduser(value), 'rb') as f:
-                    value = (os.path.basename(value),
-                             BytesIO(f.read()))
-            except IOError as e:
-                raise ParseError('"%s": %s' % (item.orig, e))
+            if(value == '-'):
```
sigmavirus24: Please remove these parentheses. It is not the style of the project and it is not typical to see in Python.

httpie/input.py
```diff
@@ -630,12 +646,15 @@ def parse_items(items,
         elif item.sep == SEP_QUERY:
             target = params
         elif item.sep == SEP_FILES:
```
sigmavirus24: This can be improved to

```py
elif item.sep == SEP_FILES and value != '-':
    # original block
```

keita1314:
@sigmavirus24 hello,According to your suggestion ,I have modified the code.
Instead of ```elif item.sep == SEP_FILES and value != '-':```
I used 
```
elif item.sep == SEP_FILES:
    if value != '-':
        # original block
    target = files
```
I think ```http -f POST file@- ```is upload an  anonymous file and I think it should be stored in ```self.args.files```
And 
```date | curl ... file@- file@- file@-``` curl will just do with the first one file@
This new commit HTTPie can works the way.

File: httpie/input.py (hunk 1 above)
Comment (keita1314) [inline httpie/input.py]: I remove this in the new comiit 343f6db

File: httpie/input.py [hunk 5]
```diff
@@ -328,6 +328,24 @@ def _parse_items(self):
...
+            if file_fields == ['']:
+                self.error(
+                    'Invalid file fileds(you just need one f@-'')'
+                )
+            if self.args.ignore_stdin or self.env.stdin_isatty:
+                self.error(
+                    'Invalid stdin(you need stdin when http -f url f@-)'
+                )
+            # only accept the first one f@-
+            self.args.files.clear()
```
sigmavirus24: What if you have

```
date | http [...] file@- file@mystuff.py file@otherstuff.py
```

keita1314: Thank you.I am trying to figure it out.

keita1314:
The commit 59d88e0 is support for multiple files

```date | http [...] file@- file@mystuff.log data@otherstuff.log```
When the file field is the same,httpie will only accept the last one,mystuff.log and otherstuff.log will be uploaded in the above example

```date | http [...] file@- data@- other@-```
The standard input will be stored in the first one file filed,'file'.
The curl works the same way.
@sigmavirus24

**Comment by sigmavirus24** [comment]
@keita1314 I haven't had a chance to try it out, but I just noticed you haven't really added any tests as part of this work. Would you mind adding some?

**Comment by keita1314** [comment]
Thank you.I am going to add some tests for this feature

**Comment by keita1314** [comment]
@sigmavirus24  add some test!

**Comment by jkbrzt** [comment]
@keita1314 Thanks for your work on this 👍

Like @sigmavirus24, I'm also thinking what the best behaviour for using the `foo@-` multiple times should be. This version, only sets the content once:

```bash
$ echo ahoy | http -vf httpbin.org/post f@-  f@-  f2@-
```

```http
POST /post HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 332
Content-Type: multipart/form-data; boundary=43a2345145e64653977da291dede8aa0
Host: httpbin.org
User-Agent: HTTPie/1.0.0-dev

--43a2345145e64653977da291dede8aa0
Content-Disposition: form-data; name="f"; filename="-"

ahoy

--43a2345145e64653977da291dede8aa0
Content-Disposition: form-data; name="f"; filename="-"


--43a2345145e64653977da291dede8aa0
Content-Disposition: form-data; name="f2"; filename="-"


--43a2345145e64653977da291dede8aa0--
```

I'm not sure if this is the best/expected behavior. I'd probably expect each of the file to have the stding data, or an error being thrown when `@-` is used multiple times. Thoughts?

sigmavirus24:
Good catch @jkbrzt 

I hadn't had a chance to test this yet but was afraid of what would happen in that case.

keita1314:
cURL
```echo ahoy | curl -XPOST  http://localhost:8000/post -H 'Content-Type:multipart/form-data' --form f=@- --form f=@- --form f2=@- --trace-ascii /dev/stdout```

```
== Info: About to connect() to localhost port 8000 (#0)
== Info:   Trying 127.0.0.1...
== Info: connected
== Info: Connected to localhost (127.0.0.1) port 8000 (#0)
=> Send header, 211 bytes (0xd3)
0000: POST /post HTTP/1.1
0015: User-Agent: curl/7.26.0
002e: Host: localhost:8000
0044: Accept: */*
0051: Content-Length: 484
0066: Expect: 100-continue
007c: Content-Type:multipart/form-data; boundary=---------------------
00bc: -------12f7256cbfbf
00d1: 
== Info: HTTP 1.1 or later with persistent connection, pipelining supported
<= Recv header, 23 bytes (0x17)
0000: HTTP/1.1 100 Continue
== Info: additional stuff not fine transfer.c:1037: 0 0
=> Send data, 484 bytes (0x1e4)
0000: ------------------------------12f7256cbfbf
002c: Content-Disposition: form-data; name="f"; filename="-"
0064: Content-Type: application/octet-stream
008c: 
008e: ahoy.
0095: ------------------------------12f7256cbfbf
00c1: Content-Disposition: form-data; name="f"; filename="-"
00f9: Content-Type: application/octet-stream
0121: 
0123: 
0125: ------------------------------12f7256cbfbf
0151: Content-Disposition: form-data; name="f2"; filename="-"
018a: Content-Type: application/octet-stream
01b2: 
01b4: 
01b6: ------------------------------12f7256cbfbf--
```
cURL also only set the content once and no error thrown.I think it is better to set the input as content only once.I have no idea why cURL treats this situation like that.I think it is a strict mode to throw error in this situation.
Thank you!
@jkbrzt @sigmavirus24

dkasak:
Was there a reason this wasn't merged or was it simply forgotten? I would really like to start using httpie, but this is preventing it for being a complete curl replacement for me.

00dani:
I was surprised to discover this doesn't already just work. `file@/dev/stdin` does but of course isn't as portable, so can we get this merged?

nglange:
Gonna give this another bump - seems like the PR is pretty stale but this and proper OAuth2 token auth are the only things keeping this from being really useful for me.

isidentical:

PR #375, opened 2015-08-23
mention httpie-jwt-auth plugin on README

hoatle:
Hello guys,

We've released https://pypi.python.org/pypi/httpie-jwt-auth and would like it to be mentioned on README so that everyone could know and use it right away.

Best regards,

Hoat Le


jkbrzt:
Thanks for sharing 👍

PR #377, opened 2015-08-31
Fix trasnsport typo in plugin manager.

mblayman:
I was reading the code and found this plugin in the method name. It worked because there was an equivalent typo in the definition.


jkbrzt:
Thanks!

Issue #381, opened 2015-09-21
--auth fails on windows


konopski:
trying to use `--auth` option on windows ends up with such a problem:
```
 http --traceback -a ext.lukasz localhost/master_service-web/rest/types
Traceback (most recent call last):
  File "C:\Python27\Scripts\http-script.py", line 9, in <module>
    load_entry_point('httpie==1.0.0.dev0', 'console_scripts', 'http')()
  File "build\bdist.win32\egg\httpie\core.py", line 104, in main
  File "build\bdist.win32\egg\httpie\input.py", line 153, in parse_args
  File "build\bdist.win32\egg\httpie\input.py", line 223, in _process_auth
  File "build\bdist.win32\egg\httpie\input.py", line 532, in prompt_password
  File "build\bdist.win32\egg\httpie\input.py", line 524, in _getpass
  File "C:\Python27\lib\getpass.py", line 95, in win_getpass
    msvcrt.putch(c)
TypeError: must be char, not unicode
```

sebastianklug:
Tried on windows 7 and could not replicate the issue.
Username: user,
pwd: passwd
httpbin.org/basic-auth/user/passwd
```
{ httpie } volvo_branch » http -a user httpbin.org/basic-auth/user/passwd
http: password for user@httpbin.org:
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 47
Content-Type: application/json
Date: Mon, 12 Oct 2015 15:28:47 GMT
Server: nginx

{
    "authenticated": true,
    "user": "user"
}

{ httpie } volvo_branch » http -a user httpbin.org/basic-auth/user/passwd
http: password for user@httpbin.org:
HTTP/1.1 401 UNAUTHORIZED
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 0
Date: Mon, 12 Oct 2015 15:29:01 GMT
Server: nginx
WWW-Authenticate: Basic realm="Fake Realm"
```

sebastianklug:
I have maked a patch for input.py
``` 
    def _getpass(self, prompt):
        #fix for https://github.com/jkbrzt/httpie/issues/381
        if is_windows:
             prompt = prompt.encode("ascii")
        #To allow mocking.
        return getpass.getpass(prompt)
```
i have also imported is_windows.

[...]

PR #382, opened 2015-09-21
[#381] --auth fails on windows

konopski:


sigmavirus24:
I'm :-1: on this. Calling a dunder method directly is very poor form.

konopski:
@sigmavirus24 what solution would you suggest then?

mdsrosa:
@konopski I think you could call `str(prompt)` instead.

jkbrzt:
Thanks :+1:

PR #384, opened 2015-09-22
Added defusedxml for outputting xml values.

kevinlondon:
I'm doing some scans of popular open source projects using [Bandit](https://github.com/openstack/bandit) and found a potential vulnerability with the xml parsing.

```
>> Issue: Using xml.etree.ElementTree.fromstring to parse untrusted XML data is known to be vulnerable to XML attacks. Replace xml.etree.ElementTree.fromstring with it's defusedxml equivilent function.
   Severity: Medium   Confidence: High
   Location: httpie/httpie/output/formatters/xml.py:46
45              try:
46                  root = ElementTree.fromstring(body.encode('utf8'))
47              except ElementTree.ParseError:
```
I addressed it by adding in defusedxml and a few tests for the xml output options, including the failure conditions. Here are the docs on [defusedxml](https://pypi.python.org/pypi/defusedxml) and the [Python docs](https://docs.python.org/3/library/xml.html#xml-vulnerabilities) talking about the potential vulnerabilities associated.

I know that introducing a new library dependency is not something to be taken lightly, so please let me know if you have any questions or suggestions.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -33,3 +33,4 @@ Patches and ideas
 * `Matthias Lehmann <https://github.com/matleh>`_
 * `Dennis Brakhane <https://github.com/brakhane>`_
 * `Matt Layman <https://github.com/mblayman>`_
+* `Kevin London <https://github.com/kevinlondon>`_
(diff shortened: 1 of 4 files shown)


jkbrzt:
Thanks a lot, @kevinlondon! Our XML formatting implementation had multiple other issues and I've removed it altogether for now.

Issue #372, opened 2015-08-04
No --pretty on Windows


nlerikheemskerk:
When using --pretty on Windows in 0.9.2 on Python 3.4.3, I get the following error:

    http: error: Only terminal output can be colorized on Windows.

This happens regardless of which --pretty option is used:

    http --pretty none http://www.google.com
    usage: http [--json] [--form] [--pretty {all,colors,format,none}]
                [--style STYLE] [--print WHAT] [--verbose] [--headers] [--body]
                [--stream] [--output FILE] [--download] [--continue]
                [--session SESSION_NAME_OR_PATH | --session-read-only SESSION_NAME_OR_PATH]
                [--auth USER[:PASS]] [--auth-type {basic,digest}]
                [--proxy PROTOCOL:PROXY_URL] [--follow] [--verify VERIFY]
                [--cert CERT] [--cert-key CERT_KEY] [--timeout SECONDS]
                [--check-status] [--ignore-stdin] [--help] [--version]
                [--traceback] [--debug]
                [METHOD] URL [REQUEST_ITEM [REQUEST_ITEM ...]]
    http: error: Only terminal output can be colorized on Windows.

Output is also by default *not* colorized.

nlerikheemskerk:
Dug a little deeper; this issue only occurs in PowerShell.

PR #385, opened 2015-10-04
Addresses issue #372

honorabrutroll:
This addresses #372 by checking whether the --output argument is used or not before raising the error.


honorabrutroll:
Forgot to delete .ropeproject leftover from vim.

PR #386, opened 2015-10-04
Addresses issue #372

honorabrutroll:
Fixes issue #372 by first checking if there is an output file being used.


jkbrzt:
Thanks :+1:

PR #393, opened 2015-10-20
Add agent url to the User-Agent

tracker1:
Add agent url for reference in the user-agent, which is a defacto standard for non-browser user agents [1]

reference: https://en.wikipedia.org/wiki/User_agent#Format_for_automated_agents_.28bots.29

NOTE: the reference is for bots, but should include automation tools such as this one.


tracker1:
An option may be to include that it's like curl or wget.

    HTTPie/%s (like curl; like wget; https://github.com/jkbrzt/httpie)

jkbrzt:
Thanks, @tracker1! I don't think it's necessary, though.

PR #394, opened 2015-10-21
Remove XML pretty-print formatting. Fixes #389.

### LINKED ISSUE #389: Maven pom.xml files become namespaced when accessed with httpie
With 

     http https://repo1.maven.org/maven2/org/apache/apache/17/apache-17.pom

the whole XML document receives a ns0 namespace. Contrast ( httpie )

    <?xml version="1.0" encoding="UTF-8"?>
    <ns0:project xmlns:ns0="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
        <ns0:modelVersion>4.0.0</ns0:modelVersion>
        <ns0:groupId>org.apache</ns0:groupId>

with ( curl)

    <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
      <modelVersion>4.0.0</modelVersion>
      <groupId>org.apache</groupId>
      <artifactId>apache</artifactId>

Comments are also lost, but not shown here for brevity

analytically:
XML parsing with Python is problematic:

- comments are by default ignored
- unable to set indent spacing in lxml
- various compability issues with ElementTree between python 2.6, 2.7 and 3.4.

Best just remove the pretty printing.


deed02392:
Can we revisit this feature? There must be a decent library for pretty printing XML, including preserving comments etc.. And even if not, it would be better to just for example compare the output length versus the input length (minus whitespace) to get a good idea on whether tidying succeeded. If it didn't just warn and show the original response body.

PR #397, opened 2015-10-24
Add = sign in assignment

zsck:


jkbrzt:
`=@` embeds the contents, but here we actually want to use `@` to simulate file upload, see:

https://github.com/jkbrzt/httpie#request-items

PR #421, opened 2015-12-15
Aitch already has an official spelling - use it

Altreus:
http://www.oxforddictionaries.com/definition/english/aitch


jkbrzt:
Great, thanks!

PR #424, opened 2015-12-28
Remove duplicate setup.py test option

t-mart:
The removed line is repeated in the method 2 statements down. I don't think both are needed.

I diff-ed the output of `python setup.py test` with and without this commit, and the output was semantically the same.


jkbrzt:
Thanks!

PR #436, opened 2016-01-28
Add XDG_CONFIG_HOME support.

Yamakaky:
On Linux and Mac, the default config directory is now "~/.config/httpie" instead
of "~/.httpie". Fixes https://github.com/jkbrzt/httpie/issues/145

### DIFF
--- a/httpie/config.py
+++ b/httpie/config.py
@@ -6,6 +6,9 @@ from httpie import __version__
 from httpie.compat import is_windows
 
 def get_default_config_dir():
+    """Returns the path to the httpie config directory.
+
+    Supports Windows and XDG_CONFIG_HOME."""
     env_config_dir = os.environ.get("HTTPIE_CONFIG_DIR")
     if env_config_dir:
         return env_config_dir
--- a/httpie/config.py
+++ b/httpie/config.py
@@ -5,13 +5,18 @@ import errno
 from httpie import __version__
 from httpie.compat import is_windows
 
+def get_default_config_dir():
+    env_config_dir = os.environ.get("HTTPIE_CONFIG_DIR")
+    if env_config_dir:
+        return env_config_dir
+    elif is_windows:
+        return os.path.expandvars(r'%APPDATA%\httpie')
+    else:
+        config_dir = os.environ.get("XDG_CONFIG_HOME",
+                                    os.path.expanduser("~/.config/"))
+        return os.path.join(config_dir, "httpie")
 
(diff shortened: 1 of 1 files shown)


Earnestly:
Make sure the path in `XDG_CONFIG_HOME` when set is absolute.  I.e. `XDG_CONFIG_HOME[0] == '/'`

Yamakaky:
If it's not, then issue an error?

Earnestly:
*Full disclaimer: I don't care about strict adherence and would rather something instead of nothing, so for example I don't care about people implementing support for `XDG_CONFIG_DIRS` until it actually matters, because usually it doesn't.*

As to your question, if the path set in `XDG_CONFIG_HOME` isn't absolute then you should use the fallback `HOME/.config`.

Yamakaky:
Done !

jleclanche:
@jkbrzt Hey, any blocker to merge this? It looks good.

skmlcd:
Any progress?

Earnestly:
Three months later, any news?

skmlcd:
@jkbrzt ?

Siilwyn:
@jakubroztocil thoughts?

Would love if this gets merged, my home directory would be thankful.

jleclanche:
@jakubroztocil Will this be merged? Will this not be merged? Can I help land it?

Siilwyn:
:point_up: same questions, if I can help by testing this locally or something else please let me know!

jleclanche:
@jakubroztocil ping - see above

aaronmehar:
@jakubroztocil Any plans to merge this at some point?

neersighted:
I would also like to see this merged. @jakubroztocil, do you have any opposition to merging as-is?

jakubroztocil:
The problem with this patch is that it changes the default config location without taking into account existing installations — people would end up with two config files one of them being be ignored (extremely bad user experience). It also assumes the existence of `~/.config/`. 

̉For this to be mergeable, it would need to:

* provide a "notify & migrate" mechanism for existing users
* have tests
* update the docs

jleclanche:
> provide a "notify & migrate" mechanism for existing users

Can just attempt to read ~/.httpie first and move it automatically if it exists right?

neersighted:
Ah, I didn't notice the lack of legacy support. I would just add another branch to the if tree, before the end, that checks if `~/.httpie` exists and just short-circuits the logic to use that folder.

Issue #423, opened 2015-12-25
`--download` reports an "incomplete download" with Content-Encoding: gzip and content-length.


twopir:
When using HTTPie to download a file which is both gzip'd and has a content-length header sent, the final output report says that the download is incomplete. This is apparently because HTTPie tracks the number of bytes it wrote, not the number of bytes it received.

Here is an example, which reproduces in HTTPie head as of about 10 minutes ago.

```$ http --debug --download GET http://torcache.net/torrent/640FE84C613C17F663551D218689A64E8AEBEABE.torrent 
HTTPie 1.0.0-dev
HTTPie data: /Users/pi/.httpie
Requests 2.6.0
Pygments 2.0.2
Python 2.7.10 (default, Oct 23 2015, 18:05:06) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.5)] darwin

>>> requests.request(**{'allow_redirects': True,
 'auth': None,
 'cert': None,
 'data': OrderedDict(),
 'files': DataDict(),
 'headers': {'Accept-Encoding': None, 'User-Agent': 'HTTPie/1.0.0-dev'},
 'method': u'get',
 'params': ParamsDict(),
 'proxies': {},
 'stream': True,
 'timeout': 30,
 'url': u'http://torcache.net/torrent/640FE84C613C17F663551D218689A64E8AEBEABE.torrent',
 'verify': True})

HTTP/1.1 200 OK
Accept-Ranges: bytes
Connection: close
Content-Encoding: gzip
Content-Length: 40082
Content-Type: application/x-bittorrent
Date: Fri, 25 Dec 2015 04:39:03 GMT
ETag: "567c9773-9c92"
Last-Modified: Fri, 25 Dec 2015 01:10:11 GMT
Server: nginx

Downloading 39.14 kB to "640FE84C613C17F663551D218689A64E8AEBEABE.torrent"
 -   0.00 %     0.00 B     0.00 B/s  -:--:-- ETA
http: error: Incomplete download: size=40082; downloaded=40260
Done. 39.32 kB in 0.22557s (174.30 kB/s)
```

Note that `size` is the value of `Content-length`, but `downloaded` is greater than `size` because of the gzip encoding.

fbidu:
Yeah, I could validate it here... I'll try to fix it

twopir:
[...]

PR #438, opened 2016-01-30
Fixes the `interrupted` property

fbidu:
Inside the `Download` class, `interrupted` evaluated to `True` if the `downloaded` size was *`different`*  than the `total_size`. This caused bug #423 where gzipped files where considered "incomplete" even though the download was successful.

This simple PR only changes the way that the `interrupted` property is evaluated - depending on `downloaded` being *less*  than total size and not just *different*

### DIFF
--- a/httpie/downloads.py
+++ b/httpie/downloads.py
@@ -272,7 +272,7 @@ class Download(object):
         return (
             self.finished
             and self.status.total_size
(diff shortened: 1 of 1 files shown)


fbidu:
This fix still leaves room for error. I'll do another fix.

PR #492, opened 2016-07-19
Test codestyle

medecau:
In this pull-request I added a test for PEP-8 conformance.

E241 and E501 error codes are currently being ignored.

I also fixed some simple codestyle errors so the new tests pass.


sigmavirus24:
I would rather see a separate test environment that just runs `pycodestyle` on the code-base instead of using an API that pycodestyle will be breaking.

medecau:
That is a very good point @sigmavirus24.

PR #497, opened 2016-08-02
Instructions for Pacman-based distributions

GuidoBR:


BoboTiG:
Thanks @GuidoBR!

Since your patch we added such section to https://httpie.io/docs#linux and we are even reworking that documentation part. Therefore I am closing the issue.

PR #506, opened 2016-08-15
Adds a 'httpie' alias to fix #505

### LINKED ISSUE #505: Add 'httpie' as a default entry point alias
This might just be me being lazy, but every damn time I try to use this tool, I type `$ httpie blah.com`, then get `-bash: httpie: command not found`, then I `$ pip install httpie` and get `Requirement already satisfied (use --upgrade to upgrade): httpie in /usr/local/lib/python2.7/site-packages`, then I have to go to this page to see that the tool is actually called `http`.

Since the repo is called `httpie`, can you please add `httpie` as an entry point alias in setup.py so that I don't get stuck in this loop any more? It'd be far more symmetrical - installing a package at one name and then executing it another is really annoying.

Just a thought,
R

Miserlou:
This is a tiny fix that adds an additional alias in the setup.py file to fix https://github.com/jkbrzt/httpie/issues/505

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,6 +83,7 @@ setup(
     entry_points={
         'console_scripts': [
             'http = httpie.__main__:main',
+            'httpie = httpie.__main__:main',
         ],
     },
     extras_require=extras_require,


sigmavirus24:
(I am not a maintainer of the project, merely another user.)

I don't think having two ways to invoke `httpie` is a good idea. It's kind of clear that the name was intentionally chosen so if you read the command `http GET https://httpbin.org/get` you're saying "aitch tee tee pee get the url `https://httpbin.org/get`". Granted it's a clever name for that reason, but it's every bit intentional. I understand your complaint in #505, but the documentation is clear. I'm personally 👎 on this.

Miserlou:
But what's wrong with having an alias? It's not an uncommon practice.

sigmavirus24:
What's wrong with defining the alias in your `<shell>rc`? That's something you can have *this second*:

```
alias httpie=http
```

And you're done

davidvhill:
I'm not a maintainer either, but +1 on the shell alias.

Miserlou:
Because of #505. It's asymmetrical to have the pypi/repo/project called 'httpie' and the client to be called 'http' with no default alias. Have the alias at the project level to correct for this.

sigmavirus24:
> But what's wrong with having an alias? It's not an uncommon practice.

"Lot's of projects (people) do it" is not a particularly good reason.

What's wrong with having the package provide two ways of doing it? There's no *need* for there to be two ways to invoke this project. If this was a brand new project and you were arguing that `http` should be `httpie` then I could get on board with that, but this is a popular, established project with documentation that very clearly points everyone to use `http`. It was chosen because of how you write the commands and how they would read. I understand that *you* have trouble with remembering that `httpie` is not a command. But the *correct* solution is not to add more technical debt to the project you chose to use. It's to add an alias for yourself locally.

And before you argue "But #505...", you filed #505. Not someone else. No one else has commented in agreement or provided any support of *your* position in that issue. And if you think this isn't technical debt, you haven't even updated the documentation examples (and yes, each example *could* be updated, but at least a representative set *need* to be updated) for this change. Maintaining documentation *is* part of maintaining this project and adding an alias is adding technical debt, even if that debt is primarily in the documentation.

And no, I won't continue this conversation with you. It's clear I'm not going to change your mind.

Miserlou:
Geez, there's no need to be a jerk about it.

hmeine:
There's a good reason for the asymmetry: The package "httpie" does not only come with the command "http", but also "https". By singling out one of them to be synonymous to "httpie", one would introduce an even worse asymmetry, in my opinion.

jakubroztocil:
Please see my comment here https://github.com/jakubroztocil/httpie/issues/505#issuecomment-646334393

Issue #496, opened 2016-08-02
Usage help says REQUEST_ITEM is required


ianpartridge:
```
$ http
usage: http [--json] [--form] [--pretty {all,colors,format,none}]
            [--style STYLE] [--print WHAT] [--headers] [--body] [--verbose]
            [--all] [--history-print WHAT] [--stream] [--output FILE]
            [--download] [--continue]
            [--session SESSION_NAME_OR_PATH | --session-read-only SESSION_NAME_OR_PATH]
            [--auth USER[:PASS]] [--auth-type {basic,digest}]
            [--proxy PROTOCOL:PROXY_URL] [--follow]
            [--max-redirects MAX_REDIRECTS] [--timeout SECONDS]
            [--check-status] [--verify VERIFY]
            [--ssl {ssl2.3,ssl3,tls1,tls1.1,tls1.2}] [--cert CERT]
            [--cert-key CERT_KEY] [--ignore-stdin] [--help] [--version]
            [--traceback] [--debug]
            [METHOD] URL [REQUEST_ITEM [REQUEST_ITEM ...]]
http: error: the following arguments are required: URL, REQUEST_ITEM
```

From the line above, REQUEST_ITEM is optional, not required.

jkbrzt:
Interesting, I don't get that message:
```
$ http
usage: http [--json] [--form] [--pretty {all,colors,format,none}]
            [--style STYLE] [--print WHAT] [--headers] [--body] [--verbose]
            [--all] [--history-print WHAT] [--stream] [--output FILE]
            [--download] [--continue]
            [--session SESSION_NAME_OR_PATH | --session-read-only SESSION_NAME_OR_PATH]
            [--auth USER[:PASS]] [--auth-type {basic,digest,ntlm,oauth1}]
            [--proxy PROTOCOL:PROXY_URL] [--follow]
            [--max-redirects MAX_REDIRECTS] [--timeout SECONDS]
            [--check-status] [--verify VERIFY]
            [--ssl {ssl2.3,ssl3,tls1,tls1.1,tls1.2}] [--cert CERT]
            [--cert-key CERT_KEY] [--ignore-stdin] [--help] [--version]
            [--traceback] [--default-scheme DEFAULT_SCHEME] [--debug]
            [METHOD] URL [REQUEST_ITEM [REQUEST_ITEM ...]]
http: error: too few arguments
```

What output does you `$ http --debug` give you?

ianpartridge:
```
$ http --debug
[...]

PR #516, opened 2016-09-01
Fix Issue #496

dongweiming:
I have same problem like https://github.com/jkbrzt/httpie/issues/496 in Python 3.

The [argparse.py#L1430](https://hg.python.org/cpython/file/3.5/Lib/argparse.py#l1430) module said:

```
if kwargs.get('nargs') == ZERO_OR_MORE and 'default' not in kwargs:
    kwargs['required'] = True
```


jkbrzt:
Thanks, @dongweiming!

PR #529, opened 2016-10-14
Added support for XSSI Vulnerability Protected JSON

yashsriv:
XSSI Vulnerability protection usually involves prefixing all JSON
data with a particular prefix to make it non-executable.
HTTPie does not display the resulting JSON in pretty print format.
My changes add support for specifying a prefix so that HTTPie can strip
it before parsing the data as JSON.

### DIFF
--- a/httpie/output/formatters/json.py
+++ b/httpie/output/formatters/json.py
@@ -17,6 +17,9 @@ class JSONFormatter(FormatterPlugin):
         ]
         if (self.kwargs['explicit_json'] or
                 any(token in mime for token in maybe_json)):
+            xssi_prefix = self.kwargs["xssi_prefix"]
+            if xssi_prefix and body.startswith(xssi_prefix):
+                body = body[len(xssi_prefix):]
             try:
                 obj = json.loads(body)
             except ValueError:
--- a/httpie/cli.py
+++ b/httpie/cli.py
@@ -223,6 +223,17 @@ output_processing.add_argument(
         ).rstrip(),
     )
 )
+output_processing.add_argument(
+    '--xssi', '-x',
+    dest='xssi_prefix',
+    metavar='PREFIX',
+    default='',
+    help="""
+    (default) Assumes incoming JSON has no XSSI Prefix
+    Allows Specifying a particular XSSI Prefix for the incoming JSON.
+
+    """
+)
 
 
 #######################################################################
--- a/httpie/output/streams.py
+++ b/httpie/output/streams.py
@@ -117,6 +117,7 @@ def get_stream_type(env, args):
                 groups=args.prettify,
                 color_scheme=args.style,
                 explicit_json=args.json,
+                xssi_prefix=args.xssi_prefix
             ),
         )
     else:


yashsriv:
Angular js supports this(See [this](https://angular.io/docs/ts/latest/guide/security.html#!#xssi)) as well as some json parsing frameworks.

jkbrzt:
@yashsriv Thanks. As far as I understand it though, this vulnerability isn't relevant to our case: HTTPie doesn't have a JavaScript runtime so *"overriding native JavaScript object constructors"* doesn't apply. 

> This attack is only successful if the returned JSON is executable as JavaScript. 

We treat JSON purely as data, and it doesn't get interpreted beyond parsing.

sigmavirus24:
@jkbrzt I think @yashsriv may have done a poor job explaining their usecase. As I understand their change, they're trying to communicate with a service that returns XSSI protected JSON. Now, with today's HTTPie they lose JSON syntax highlighting because of this. If they were able to specify the prefix, however, that would be stripped, then they could see the JSON as they would see normal JSON via HTTPie.

While I agree with you that this doesn't strengthen HTTPie's security posture, I think this serves to help its users of these types of APIs.

jkbrzt:
@yashsriv Now I see what this is about — I have done a poor job reading the description the first time around. Thanks @sigmavirus24.

 So I understand that `")]}',\n"` is the de facto standard prefix.  

What if we simply tweaked HTTPie's JSON handling so that when JSON is expected but the content is not valid, the parser would try again while skipping anything before the first `[` or `{`? And then, if the remainder turns out to be valid JSON, it would format it and print out `prefix` + `formatted_json`.

Do you have an example URL?

yashsriv:
@jkbrzt I don't have an example URL because I encountered this issue when I implemented XSSI protection on a REST api I'm working on for a private project. Also, I added the option to specify a prefix because I read online that some services use `while(1);` and other prefixes too. So, creating a generalised handler seems difficult. In my way, if you encounter an api with a different prefix, you just specify the prefix and it works.

yashsriv:
This is the difference on my machine:

img-json-example

adarshaj:
I'm interested in having this feature. Would like to see the feature merged! Is there anything which can be improved to get this done? 
/cc @jkbrzt

httpie/output/formatters/json.py
```diff
@@ -17,6 +17,10 @@ def format_body(self, body, mime):
         ]
         if (self.kwargs['explicit_json'] or
                 any(token in mime for token in maybe_json)):
+            if 'xssi_prefix' in self.kwargs:
+                token = self.kwargs["xssi_prefix"]
+                if body[0:len(token)] == token:
```
sigmavirus24: Would it not be simpler to do:

```py
if body.startswith(token):
```

It looks as though `token` being `''` won't get through to here because otherwise you'll always find this condition to be true and you'll always make a copy.

That said, even if you specify the common value that @jkbrzt found, this should be a relatively cheap check (to use startswith).

yashsriv:
Rebased and updated as per @sigmavirus24 's recommendation

fearphage:
Is this waiting on anything else?

adarshaj:
I'm still having to fallback to  `http $URL | tail -n1 | jq` .. can this diff be pls merged? :pray: 
/cc @jakubroztocil

sakshamsharma:
Bump. Can we hope that this PR gets merged soon?
/cc @jakubroztocil

httpie/output/formatters/json.py
```diff
@@ -17,6 +17,10 @@ def format_body(self, body, mime):
         ]
         if (self.kwargs['explicit_json'] or
                 any(token in mime for token in maybe_json)):
+            if 'xssi_prefix' in self.kwargs:
```
fearphage: This will always be true since you defaulted it in the cli arguments. I think it'd be better to reverse this logic a bit:

```python
xssi_prefix = self.kwargs["xssi_prefix"]

if xssi_prefix and body.startswith(xssi_prefix):
  body = body[len(xssi_prefix):]
```

BoboTiG:
Done with #1130.

Thanks for the original patch @yashsriv :pie:

Issue #559, opened 2017-02-09
Allow --verify=False and --verify=True (convert to no/yes) or raise warning on this bad input


hangtwenty:
I went down a rabbit hole trying to "fix" my OpenSSL and/or PyOpenSSL again, as I've had to deal with in the pass... only to realize the reason nobody else was seeing my issue is that it was completely my fault, bad usage. But maybe there is an enhancement opportunity here.

Compare these brief outputs (I will put `--debug` below):

```
$ http --verify=no "https://github.com" | head -n5


<!DOCTYPE html>
<html lang="en" class="">

$ http --verify=False "https://github.com" | head -n5

http: error: Error: [('system library', 'fopen', 'No such file or directory'), ('BIO routines', 'BIO_new_file', 'no such file'), ('x509 certificate routines', 'X509_load_cert_crl_file', 'system lib')]
```

I actually tried `--verify=False` first ... for some reason didn't try `--verify=no` until going through all sorts of other stuff. Maybe we can catch it and refuse to proceed, or at least print a warning about the likely error that will follow. Here's the full `--debug` output:

```
$ http --verify=False --debug "https://github.com"
HTTPie 0.9.9
Requests 2.13.0
Pygments 2.2.0
Python 2.7.12 (default, Oct 28 2016, 12:57:16)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)]
/usr/local/opt/python/bin/python2.7
Darwin 16.4.0

<Environment {
    "colors": 256,
    "config": {
        "__meta__": {
            "about": "u'HTTPie configuration file'",
            "help": "u'https://github.com/jkbrzt/httpie#config'",
            "httpie": "u'0.9.6'"
        },
        "default_options": "[]"
    },
    "config_dir": "/Users/mfloering/.httpie",
    "is_windows": false,
    "stderr": "<open file '<stderr>', mode 'w' at 0x10754d1e0>",
    "stderr_isatty": true,
    "stdin": "<open file '<stdin>', mode 'r' at 0x10754d0c0>",
    "stdin_encoding": "UTF-8",
    "stdin_isatty": true,
    "stdout": "<open file '<stdout>', mode 'w' at 0x10754d150>",
    "stdout_encoding": "UTF-8",
    "stdout_isatty": true
}>

>>> requests.request(**{
    "allow_redirects": false,
    "auth": "None",
    "cert": "None",
    "data": {},
    "files": {},
    "headers": {
[...]

PR #560, opened 2017-02-10
Turn --verify=False/True to --verify=no/yes

hangtwenty:
One way to address #559 -- instead of warning or throwing an error, just accept "True" and "False"
as synonyms of yes/no

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -159,8 +159,10 @@ def get_requests_kwargs(args, base_headers=None):
         'data': data,
         'verify': {
             'yes': True,
-            'no': False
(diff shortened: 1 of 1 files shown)


jkbrzt:
@hangtwenty thanks for the pull request. I can understand the confusion.  Here's a possibly more robust solution. 

https://github.com/jkbrzt/httpie/blob/de38f86730bae524b9cbec317a6275a74987400a/httpie/client.py#L163

```python
        'verify': {
            'yes': True,
            'true': True,
            'no': False,
            'false': False,
        }.get(args.verify.lower(), args.verify),
```

PR #561, opened 2017-02-14
Add the packaging metadata to build the httpie snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be continuously delivered to many users through the Ubuntu store.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,4 +1,4 @@
-name: httpie
+name: http
 version: master
 summary: Modern command line HTTP client
 description: |
@@ -15,7 +15,6 @@ apps:
   http:
     command: http
     plugs: [network, home]
-    aliases: [http]
 
 parts:
   httpie:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,23 @@
+name: httpie
+version: master
+summary: Modern command line HTTP client
+description: |
+  HTTPie (pronounced aitch-tee-tee-pie) is a command line HTTP client. Its goal
+  is to make CLI interaction with web services as human-friendly as possible.
+  It provides a simple http command that allows for sending arbitrary HTTP
+  requests using a simple and natural syntax, and displays colorized output.
+  HTTPie can be used for testing, debugging, and generally interacting with HTTP servers.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  http:
+    command: http
+    plugs: [network, home]
+    aliases: [http]
+
+parts:
+  httpie:
+    source: .
+    plugin: python


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/httpie
    $ cd httpie
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

(dangerous because the snap you build yourself doesn't come signed from the trusted store. Once you push it to the store, your users will just have to do $ sudo snap install http)

If you have any questions or comments, please let me know.

sigmavirus24:
Hi, httpie user and repository watcher here,

Why does this belong in the repository here? Debian, for example, keeps a copy of the source repo and stores their packaging information there. RHEL/Fedora/etc. all do that in separate repositories. Further, are snaps really that popular/widely used? My impression is that they aren't and that there are still not-insignificant security concerns with the format. Finally, if this is a secure distribution model, couldn't they have used Let's Encrypt for snapcraft.io? You'd think they'd be using HTTPS by now.

elopio:
Hello @sigmavirus24, thanks for taking a look.

The snapcraft metadata could live in a separate repo. However, the process and the tools are optimized to make it as lightweight as possible, so it requires little maintenance and gives you some niceties when it's in the same upstream repo as the source code. For example, you could easily hook the repo with travis or launchpad [1] to push a new version to the edge channel in the store every time master changes, which is a nice way to get early feedback before releasing a new stable version. And those users on the edge channel will get an automatic update every time upstream pushes something new.

snaps are being used extensively by people in Ubuntu 14.04, 16.04, 16.10 and the upcoming 17.04. That's a lot of people, and we have work in progress to improve the experience in the other distros [2], which will increase the adoption there too. The huge thing here is that you have to push only one version, and it will work the same everywhere. If the push is automated as mentioned above, then the required work is very little and the benefits are plenty.

About the website, I see the bug has already been reported: https://github.com/ubuntudesign/snapcraft.io/issues/115 Your comment in there would be appreciated, and I'm sure it will help the web team to prioritize it for a fix in the following version.

Can you please tell me about those security concerns you mention? We are constantly hardening the operating system with every new release, so I'd like to see what's the status of the things that worry you.

[1] https://snapcraft.io/docs/build-snaps/ci-integration
[2] https://github.com/snapcore/snapd/wiki/Distributions

sigmavirus24:
> Can you please tell me about those security concerns you mention?

http://www.pcworld.com/article/3063134/linux/ubuntus-snap-packages-arent-yet-as-secure-as-canonicals-marketing-claims.html

I realize now that it's not Snap but X11. Apologies. I just remembered the security claims didn't seem as impressive as Canonical was claiming when it first came out.

> And those users on the edge channel will get an automatic update every time upstream pushes something new.

This project's development has slowed down quite a bit. I'm not sure how valuable that would be.

> snaps are being used extensively by people in Ubuntu 14.04

In 14 days the massive user base of 14.04 has picked up snaps? Do you have actual metrics for that?

---

Anyway, this is a tiny file. I still don't like my downstream redistribution packaging in the upstream repository unless the author is the packager, but that may just be me.

elopio:
right, it's hard to convince people to stop using X11. But that's how it all began. Unity, Mir and a secure environment for installing apps in the ubuntu phone. Snaps should be able to use x11, mir, wayland, or whatever is running on the system, and be as secure as possible. Even if you have X11, it's still safer than if you just use it in a classic OS. The full snappy version, Ubuntu Core, won't have any of those problems, but any way, this is a command-line tool, so it won't have any of those problems.

About the speed, even if you don't land new features every day, it's pretty useful to have the whole delivery pipeline automated. What's important is this:

http://packages.ubuntu.com/trusty/httpie 0.8.0-1
http://packages.ubuntu.com/xenial/httpie 0.9.2-1

httpie has made a lot of improvements since then, and before snaps they were not easily accessible to the users in trusty and xenial. We will have many users in xenial until 2021. Even if you land one feature every month, in 5 years that's a lot. That made us realize a long time ago that the Ubuntu schedule is great to provide a stable and secure base system, but not good for most of the upstreams; so we started investing piece by piece on improving it and now have the technology to basically get out of the way and let developers deliver their software directly to users, without breaking the system.

The thing here is that there is no packager, because packaging is done in a couple of minutes (the time it took me to fill the fields in this snapcraft.yaml). There is no maintainer, because changes are confined to a secure space so your CI system does all the uploads without manual intervention. This tiny file is all that's required to get something like like pip+virtualenv+automatic updates+rollbacks+security, so why not?

About 14.04 stats, I don't have yet a way to know where the downloads come from. The store team started with a basic set of stats, and are adding more as they are being requested by upstreams. We are in the phase of early adopters: http://elopio.net/blog/call-for-testing-trusty-snaps/ and my estimates after working with those early adopters is that some projects are getting around 200 installs per day, 10% of which are from trusty. But until the next cycle of stats is deployed, it's just a guess. 

What I can tell you for sure is that httpie already has 3705 snappy users. We needed to call http to run some tests, and the process wasn't smooth enough at that time to ask you to push it to the store. So a friend made the package but didn't publicize it in any way, these are hardcore early adopters that have found it by themselves after exploring a lot; they are the ones who report bugs and help us test our new releases before we give them to everybody. What we want now is to transfer it to you, to get your feedback about the store and improve it based on your real needs. If you put in your install instructions the option to install httpie from a snap we can actually see how the stats increase and how many of your users prefer that option, and then you will have plenty of information about how the addition of a snap impacted your project.

I'm sorry, I have written a lot more than what I planned, I'll stop now. I get excited because I'm enjoying this very much. This is a great job, I often stumble with people who writes the basic tools that let me make my work, and can have a nice chat with them. I have taught my team to appreciate pretty python code by not letting them land anything otherwise: https://github.com/snapcore/snapcraft/blob/master/runtests.sh#L52 @sigmavirus24 thanks a lot for that :smile:

elopio:
Ping. Anybody had a chance to review this?

elopio:
Ping again, I'm trying to catch up with old PRs.
Somebody uploaded http to the store, and it already has more than 3000 users.
We prefer upstream than maintainers, so if you want to take ownership of the snap at any time in the future, just let me know.

We also have a new service that makes the maintainership an effortless task. In https://build.snapcraft.io you can configure the automated build and delivery. Take a look!

pura vida

snap/snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+apps:
+  http:
+    command: http
+    plugs: [network, home]
+
+parts:
+  httpie:
+    source: .
+    plugin: python
```
BoboTiG, line 22: As there is more `requirements.txt`, I am wondering it this is missing?
```suggestion
    plugin: python
    python-packages:
      - "."
```

snap/snapcraft.yaml
```diff
@@ -0,0 +1,22 @@
...
+  requests using a simple and natural syntax, and displays colorized output.
+  HTTPie can be used for testing, debugging, and generally interacting with HTTP servers.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  http:
+    command: http
+    plugs: [network, home]
```
BoboTiG, line 17:
```suggestion
  http:
    command: http
    plugs: [network, home]
  https:
    command: https
    plugs: [network, home]
```

BoboTiG:
Closing this one as I am working on an up-to-date version. Thanks @elopio, I'll not forget to give you credits :)

Issue #524, opened 2016-10-01
Colorized output doesn't follow terminal ANSI color styles.


winneon:
HTTPie prefers [solarized](http://ethanschoonover.com/solarized) for its color palette, and it seems it doesn't properly follow whatever terminal style / colors that the end-user has set.

For example: I have specific ANSI colors set in my terminals, whether it be Terminal.app or HyperTerm on macOS, and yet, HTTPie still prefers solarized.

Terminal.app:

![](https://i.imgur.com/3OlURws.png)
![](https://i.imgur.com/A2GleZ0.png)

HyperTerm:

![](https://i.imgur.com/2FHgm2y.png)
![](https://i.imgur.com/EWyrpla.png)

jkbrzt:
@winneon There currently is no support for this. HTTPie uses color schemes from Pygments, the only custom theme is our [solarized](https://github.com/jkbrzt/httpie/blob/0913e8b2ef80ad0cd96174ffff14e0e7d3f931a1/httpie/output/formatters/colors.py#L172-L241). There's the `--style` option by which can choose a theme:

```
  --style STYLE, -s STYLE
      Output coloring style (default is "solarized"). One of:

          algol, algol_nu, autumn, borland, bw, colorful, default,
          emacs, friendly, fruity, igor, lovelace, manni, monokai,
          murphy, native, paraiso-dark, paraiso-light, pastie,
          perldoc, rrt, solarized, tango, trac, vim, vs, xcode

      For this option to work properly, please make sure that the $TERM
      environment variable is set to "xterm-256color" or similar
      (e.g., via `export TERM=xterm-256color' in your ~/.bashrc).
```

What I believe could be done though is to create a custom theme whose colors would be dynamically assigned from the terminal ANSI color scheme.

winneon:
That would be great! I'm not able to make a pull request at the moment because my PC's motherboard is out for replacement, but if there isn't a change implemented when I get it back, then I will take a look into that.

dsego:
When I just change the formatter to `self.formatter = TerminalFormatter()`, without any styles I get the correct colors from my iTerm profile (material design in my case, see the screenshot). Weird enough, using Terminal256Formatter with default styles results in ugly colors. I propose to set to TerminalFormatter by default and enable the other one only when the style flag is set.


[...]

PR #568, opened 2017-03-10
Follow terminal ANSI color styles

dsego:
Removes the default solarized color scheme and custom http lexer class.

### DIFF
--- a/httpie/cli.py
+++ b/httpie/cli.py
@@ -20,7 +20,7 @@ from httpie.input import (
     PRETTY_STDOUT_TTY_ONLY, SessionNameValidator,
     readable_file_arg, SSL_VERSION_ARG_MAPPING
 )
-from httpie.output.formatters.colors import AVAILABLE_STYLES, DEFAULT_STYLE
+from httpie.output.formatters.colors import AVAILABLE_STYLES, DEFAULT_STYLE, PRESET_STYLE
 from httpie.plugins import plugin_manager
 from httpie.plugins.builtin import BuiltinAuthPlugin
 from httpie.sessions import DEFAULT_SESSIONS_DIR
--- a/httpie/cli.py
+++ b/httpie/cli.py
@@ -20,7 +20,7 @@ from httpie.input import (
     PRETTY_STDOUT_TTY_ONLY, SessionNameValidator,
     readable_file_arg, SSL_VERSION_ARG_MAPPING
 )
-from httpie.output.formatters.colors import AVAILABLE_STYLES
+from httpie.output.formatters.colors import AVAILABLE_STYLES, DEFAULT_STYLE
 from httpie.plugins import plugin_manager
 from httpie.plugins.builtin import BuiltinAuthPlugin
 from httpie.sessions import DEFAULT_SESSIONS_DIR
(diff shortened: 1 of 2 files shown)


jakubroztocil:
Thanks, @dsego!

I've tested it with a couple of color schemes in iTerm and the vanilla `TerminalFormatter` seems to work quite well. Colored terminal output is generally tricky, though, so It will require more extensive testing on different platforms/configurations before this can become the new default. 

So instead of making it the default behaviour now, I'd go for a transitional period when this formatter is available only via an extra style option: `--style=native`.

As for the `Solarized` style, I'd keep it around — it's quite popular and removing it would break  backward compatibility (hopefully it will be included in Pygments one day).

Also, what were your thoughts behind removing the custom HTTP lexer? It provides a nicer formatting for HTTP headers than the default one so unless there's a good reason, let's keep it as well. 

Context: #524

dsego:
Hey, thanks for the review @jakubroztocil. 

I agree it might be best to have it  is as an extra style option. 
As for the custom lexer, I found that the one included with pygments does really well with the vanilla `TerminalFormatter`,  http headers show in nice colors with good contrast. 

Here is a comparison of the default & custom lexer with [Snazzy theme](https://github.com/sindresorhus/iterm2-snazzy) in iTerm2:

headers

It may be the case that some color schemes supplied with pygments are not handling http headers nicely (monokai works well for example), so I agree with keeping the custom lexer.

jakubroztocil:
The Snazzy theme looks great  :) 

> It may be the case that some color schemes supplied with pygments are not handling http headers nicely (monokai works well for example), so I agree with keeping the custom lexer.

We could also simply not apply the custom HTTP lexer when the "native" theme is active.

dsego:
Since _native_ and _default_ are taken what do you think about one of the following:
_profile_, _auto_, _preset_, _local_, _terminal_, _term_, _scheme_?

jakubroztocil:
Thanks, @dsego! I'll polish it a bit and it include in the next release.

9mm:
@dsego what font is that in your screenshot?

dsego:
@9mm I believe this was roboto mono.

9mm:
Awesome thanks. How did you get those colors to appear like that? I installed the iterm2 theme and every --style option I try looks horrible. I'm thinking of making my own but not sure how

dsego:
@9mm You may want to check out this issue https://github.com/jakubroztocil/httpie/issues/524.

9mm:
Oh I see. Thank you. What might be the best way to change this if I use homebrew and a new update will just clobber it?

dsego:
@9mm The code in this PR is already merged in. So the next release should probably have an additional _--style_ option to match the colors to whatever iTerm2's color scheme is. When and how that happens is left to be seen.

9mm:
Cool, thanks for explaining

ThiefMaster:
This change is bad :(

I use the default colors in the terminal and this is fine for pretty much everything - but now HTTPie uses an awful dark blue (on the black background almost everyone has in their terminal) for some parts, which is very hard to read.

Is there some global config file I can create to get back the old style? (and no, I don't want to use shell aliases for that)

jakubroztocil:
You select the old default colors (< v1.0.0)  with `--style=solarized`.  To make it permanent, add it to `default_options` in your confit file. https://httpie.org/doc#config

PR #594, opened 2017-07-11
Use 'UTF-8' as default encoding

CzBiX:
Requests never return `None` when charset is missing in content type. So the code at L83 wouldn't work as expected.

You can use this command to test it: `http https://www.google.com/intl/zh-CN/policies/`.
It should output Chinese at end after apply this patch:
image

### DIFF
--- a/httpie/models.py
+++ b/httpie/models.py
@@ -80,7 +80,9 @@ class HTTPResponse(HTTPMessage):
 
     @property
     def encoding(self):
-        return self._orig.encoding or 'utf8'
+        # requests use 'ISO-8859-1' as fallback, but we choose 'UTF-8'.
+        # see discussion in https://github.com/requests/requests/issues/2086
+        return self._orig.encoding if 'charset=' in self.content_type else 'utf8'
 
     @property
     def body(self):


sigmavirus24:
> Requests never return None when charset is missing in content type.

This is a bold claim and if true, a bug in Requests. Please provide a bug report there as well as something that can be used to reproduce this.

sigmavirus24:
You no longer need to file a bug. I looked at the code in Requests and found the [utility function](https://github.com/requests/requests/blob/649efc8c9b9dbc83d707d1d161f2c18f86d3b257/requests/utils.py#L410) that generates the encoding attribute of the Response. It only returns `ISO-8859-1` when the content-type is a `text/*` mimetype. This is as specified. As it stands, this patch is actively wrong because Requests does return `None`. There *may* be times where Requests does the right thing that a user of HTTPie wants to override. In those cases, I suggest a new flag for the user to tell HTTPie what to do. I don't think you should ignore the encoding here regardless of what it *should* be detected as because a subset of possible websites don't properly specify a charset.

CzBiX:
Oh, there is a inaccuracy in my description. It should be:

> Requests never return None when charset is missing in text content type.

Please take a view of my patch, and think about results with/without this patch when content type missing charset info:
1. content type is `text/*`, requests: `ISO-8859-1`, httpie(before/after): `ISO-8859-1`/ `UTF-8`
2. content type is others, requests: `None`, httpie(before/after): `UTF-8`/ `UTF-8`

If a website use charset other than `ISO-8859-1`/`UTF-8` and missing charset, httpie can't decode text properly too, because we don't find meta tag in html or use `chardet` to detect charset(which is slow, requests only use it in `Response.text` property).

Here is my point:
1. This patch will not ignore the encoding in http header, but return `UTF-8` if missing charset only.
2. If content charset is missing in header and isn't `ISO-8859-1`/`UTF-8`, we can't decode content properly.
3. I think my patch is acceptable in most of situation, but we should add an option to force content charset, and it's belongs to another RP.

sigmavirus24:
> because we don't find meta tag in html or use chardet to detect charset(which is slow, requests only use it in Response.text property)

Requests has utilities for finding the meta tag in the HTML. That's plausible and a better solution than this.

Requests, to be clear, *only* uses `chardet` when you access `Response.apparent_encoding` which *might* be used in `Response.text`.

> I think my patch is acceptable in most of situation

I don't entirely agree, but I think that was already apparent. 😄

CzBiX:
`http https://www.google.com/intl/zh-CN/about/` still output garbage characters.
@jakubroztocil  It's possible to merge this? It's very annoying to non-ASCII users.

dlukes:
FWIW, `curl` seems to assume `UTF-8`:

```sh
$ curl -X POST "https://lindat.mff.cuni.cz/services/translation/api/v2/models/en-cs" -d input_text="All humans are born equal."
Všichni lidé jsou si rovni.
$ http -pb -f post "https://lindat.mff.cuni.cz/services/translation/api/v2/models/en-cs" input_text="All humans are born equal."
VÅ¡ichni lidÃ© jsou si rovni.
```

If you don't think this is a good option for `httpie`, @jakubroztocil, could you please state why and close this PR, so that people know it won't be happening? Otherwise, it would be great if you could merge it if/when you get the chance :)

If you're worried about crashing with a `UnicodeDecodeError` when the data happens to *not* be `UTF-8`, the PR could be extended to use the `replace` error handling strategy, which is resilient to this type of problem:

```python
>>> b'\x9a\x9dastn\xe1 ko\xe8ka\n'.decode("utf-8", errors="replace")                                                                                                                                                                                                                        
'��astn� ko�ka\n'
```

Almad:
This has broader considerations—see #1022

BoboTiG:
Superseded by #1110.

BoboTiG:
Thanks for the patch anyway @CzBiX :pie:

PR #595, opened 2017-07-12
Use ` unicode` to know if Python 2 or 3

cclauss:
This passes linter tests like flake8


sigmavirus24:
This is undesirable for a few reasons:

1. This project doesn't use Flake8, so its "linter tests" aren't important
2. This assigns `unicode` to `str` and then `str` to `bytes` which means that `isinstance(bytes(), unicode)` will be `True`.
3. The existing code is more readable.
4. If the project were to update this, it would be better to rely on `six`.

PR #607, opened 2017-09-06
Add 'random' color styles option

parhamfh:
I find it easier to distinguish between different calls in the terminal when the output is colored differently for each call, so you can more easily tell when one ends and the other begins. So I added `random` as an option!
Locally I rotate between only 3 of the colors since I don't like all of them but didn't want to make that choice for all.

### DIFF
--- a/demo_config.json
+++ b/demo_config.json
@@ -1,0 +1,10 @@
+{
+    "__meta__": {
+        "about": "HTTPie configuration file",
+        "help": "https://github.com/jkbrzt/httpie#config",
+        "httpie": "0.9.6"
+    },
+    "default_options": [
+        "--style=random"
+    ]
+}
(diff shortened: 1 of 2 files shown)


isidentical (changes requested):
I don't think httpie is the right place to make this change. You can easily create a `pygments` plugin that installs a `random` style (as you see the `AVAILABLE_STYLES` is dynamically determined, so if you install your plugin through `pygments`'s entry points you can use that style in HTTPie), and return a random lexer every time it is called. Or you can create an HTTPie formatter plugin to do so (but I think pygments would be way easier for you, since it has a stable interface). @jakubroztocil any comments on whether we should have this or not?

parhamfh:
That sounds good @isidentical! If there are ways around it that is cool, this was just an idea I had back when I opened this. I would imagine it is obsolete as well so I'll close it! 😅

Issue #150, opened 2013-07-23
Add option to ignore stdin data


zapman449:
```
http: error: Request body (from stdin or a file) and request data (key=value) cannot be mixed.
```

This error is rather cryptic.  I'm writing a shell script to interact with a rest API, and I'm doing this:

```
http --output subnet --pretty format --form post "${URLHOST}subnets" "cidr=${cidr}" "name=${name}" "description=${name}" "availabilityZoneId=${az}" "networkId=${VPCid}" "$ACCESSKEY" "$SECRETKEY" "$ACCOUNTID" "_providerId=aws" "_regionId=${REGION}"
```

Which produces (when preceded by 'echo' a command that works perfectly well.  However, this version gives the error in the subject.

Am I being stupid, or is this something else?

FYI: The result of the above is as followed (sanitized slightly)

```
http --output subnet --pretty format --form post https://test.grid.domain.com/subnets cidr=10.124.33.128/25 name=Load_balancers description=Load_balancers availabilityZoneId=us-east-1c networkId=vpc-31433e7e x-gridauth-accesskey:DOYOUNEEDTOKNOW x-gridauth-secretaccesskey:WHATISWITHALLTHEQUESTIONS x-gridauth-accountid:11223344556677 _providerId=aws _regionId=us-east-1
```

CrazyPython:
This is very confusing, and goes against httpie being able to work anywhere. Using such a command inside a bash loop is a very common application. Httpie should not complain when key/value is given and it receives stdin: instead letting key/value dominate.

sigmavirus24:
> Httpie should not complain when key/value is given and it receives stdin: instead letting key/value dominate.

That maybe is intuitive to you, but to many others the exact opposite is far more intuitive. Because tooling should not try to guess, it is far better to raise an error and provide you an escape hatch if you feel you know better, which this does.

jenstroeger:
I ran into this issue when using httpie inside of a function passed to [GNU parallel](https://www.gnu.org/software/parallel/):
```bash
function call_api {
    local FOO=`http --json --auth $TOKEN: post $HOST/api/foo name="Ţẽṧẗ" < /dev/tty | python3 -c "import json,sys;obj=json.load(sys.stdin);print(obj['result']['id']);"`
    […]
}

export -f call_api
export HOST=…
export TOKEN=…
parallel --env HOST --env TOKEN ::: call_api call_api call_api call_api call_api
```
[...]

PR #630, opened 2017-11-13
Clarify error message

CrazyPython:
#150 was addressed by adding `--ignore-stdin`, but the error message wasn't clarified to mention that option.


jakubroztocil:
Thanks!

PR #633, opened 2017-11-18
fix env version attribute

darshanime:
if multiple versions of `httpie` are installed on a machine, the `env.config.__meta__.httpie` attribute doesn't have the current version from `__init__.__version__`
this can cause problems if the `env` object is used to make decisions to enforce compatibility based on the httpie version.


jakubroztocil:
Thanks!

PR #638, opened 2017-12-07
Update CONTRIBUTING.rst to include correct Makefile targets.

gtback:
Hi, I'm just getting started with developing on HTTPie. I was going
through the contributing guide and noticed that a couple of the `make`
commands seem a little stale.


jakubroztocil:
Thanks!

PR #641, opened 2017-12-21
Fixed a sentence in the README regarding passing arbitrary request data

specious:
It didn't make sense when I read it.


jakubroztocil:
Thanks, @specious!

PR #648, opened 2018-01-18
Note about re-ordering when using colored output

iannickb:


jakubroztocil:
Thanks, @iannickb. I'm sure this could be communicated better in the docs, but the sorting is mentioned there:

> Also, the following formatting is applied: HTTP headers are sorted by name. JSON data is indented, sorted by keys, and unicode escapes are converted to the characters they represent.
> https://httpie.org/doc#colors-and-formatting 

Btw, the sorting is the result of `--pretty=format`, not `--pretty=colors` (both are implied with `--pretty=all` which is the implicit default for terminal output).

PR #663, opened 2018-03-14
Support using styles from Pygments plugins

psliwka:
`pygments.styles.STYLE_MAP` contains only styles built directly into Pygments library. To list all available styles (including styles registered by plugins), one should use `get_all_styles` generator.

For respective Pygments documentation, see: http://pygments.org/docs/styles/#getting-a-list-of-available-styles


psliwka:
@jakubroztocil: ping ;)

jakubroztocil:
Thanks Piotr!

PR #664, opened 2018-03-21
Use parentheses in describing sessions

tbodt:
It's a little more readable in this case than using em dashes.


jakubroztocil:
Thanks!

PR #683, opened 2018-06-08
[#681] fixed https certificates for tests;

### LINKED ISSUE #681: Fix Travis / tests
https://travis-ci.org/jakubroztocil/httpie/jobs/385658131#L598-L616

yevgenys:
fixes #681


jakubroztocil:
@yevgenys thanks for the pull request. The solution is not exactly correct (it reverses the logic of the trusted/untrusted states) but it has helped me to come up with the 749b1e2aca fix  👍

PR #698, opened 2018-09-03
fixed wrong value in bash command

0xflotus:
It should be 'value' not 'foo' to get the corresponding output


0xflotus:
Ok, I will fix the output

Issue #566, opened 2017-03-08
Brew installed version doesn't work with plugins


rshurts:
Brew is listed as the recommended way of installing httpie on macOS. However, it doesn't work with auth plugins.

For example, if you `pip3 install requests-hawk` and then run `http --help` hawk will not show as an auth type. If you `pip3 install httpie-oauth` it will install httpie via pip as a dependency and overwrite the brew installed link in `/usr/local/bin/http` and now all the plugins will show because it isn't using the brew installed version.

I suggest changing the documentation to read `pip3 install httpie` as the recommended method of installing on macOS.

jkbrzt:
@rshurts you're right. Plugin installation for Homebrew-based installs doesn't work out of the box. 

There are some advantages to keeping Homebrew as the recommended method, though. Mainly it makes updates really easy and the package can depend on Python 3 (i.e. much better SSL support, etc.). 

The ideal world solution would be to make all plugins available as Homebrew packages (e.g. `brew install httpie-oauth`, etc.) . But that would mean additional work for maintainers so not all plugins would end up in Homebrew. 

Using `pip` to install plugins looks like the easier way to go. To make it work with Hombrew, it would mean instructing `pip` to use  (currently) `/usr/local/Cellar/httpie/0.9.8_2/libexec/lib/python3.6/site-packages` as the installation directory. 

But plugins installed there won't survive HTTPie upgrades. 

Perhaps it would be possible to tweak the [Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/httpie.rb) to make it possible to load installed plugins from the Homebrew Python 3 `site-packages` (or from another location). 

…or some sort of a custom plugin installer.

rshurts:
In the meantime, maybe a note in the documentation around plugins saying "brew doesn't work with plugins and to pip install httpie for plugin functionality" would save some headaches.

aheissenberger:
There is also a problem with the read rights when user has no admin rights:
```bash
$ http --help
Traceback (most recent call last):
  File "/usr/local/bin/http", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/local/Cellar/httpie/1.0.0/libexec/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3123, in <module>
    @_call_aside
  File "/usr/local/Cellar/httpie/1.0.0/libexec/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3107, in _call_aside
[...]

PR #701, opened 2018-09-06
Add extra_site_dirs config option to solve #566

matthew16550:
Hi, it's my first contribution to HTTPie.  Thanks for making a really great tool!

This is a possible solution for #566 by letting users specify extra directories for `sys.path`.


httpie/config.py
```diff
@@ -84,7 +84,8 @@ class Config(BaseConfigDict):
     about = 'HTTPie configuration file'
 
     DEFAULTS = {
-        'default_options': []
+        'default_options': [],
+        'extra_site_dirs': [],
```
matthew16550: We could detect when running in a Homebrew installation and set `extra_site_dirs` here to a useful default such as the example in `README.rst`.  I'm not sure if it's a good idea.

.travis.yml
```diff
@@ -61,7 +61,11 @@ install:
       if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
           if [[ -n "$BREW_PYTHON_PACKAGE" ]]; then
               brew update
-              brew install "$BREW_PYTHON_PACKAGE"
+              if ! brew list --versions "$BREW_PYTHON_PACKAGE" >/dev/null; then
```
matthew16550: @jakubroztocil FYI this seems to fix the Travis builds that have been broken for a few days

matthew16550: Figured this is better in a separate PR so moved to #705.

jakubroztocil:
@matthew16550 thanks for the PR. I think it's a really good idea. I'll review it thoroughly when i have the time.

matthew16550:
Hi @jakubroztocil is this still relevant?  I haven't been paying any attention here.

If yes then I'd be happy to update the PR.

jakubroztocil:
@matthew16550 thanks for the PR and for coming back! Yes, it is still relevant. Let me look through the old solution in the coming days and share my thoughts.

isidentical:
Thanks for your contribution @matthew16550, we decided to solve this problem similarly but in a different fashion in #1200. Closing in favour of that!

Issue #699, opened 2018-09-03
Feature Request: more user friendly message for download time elapsed


hanxue:
When a `--download` operation takes a long time, the output will be 

`Done. 51.05 MB in 1653.26288s `

When the number of seconds is in the thousands, it requires more effort to calculate how many minutes of hours elapsed for the download operationg. 

A full example:

```
$ http --download http://dl.caffe.berkeleyvision.org/bvlc_googlenet.caffemodel
HTTP/1.1 200 OK
Accept-Ranges: bytes
Connection: keep-alive
Content-Length: 53533754
Content-Type: application/octet-stream
Date: Mon, 03 Sep 2018 07:23:08 GMT
ETag: "54934a20-330dc3a"
Last-Modified: Thu, 18 Dec 2014 21:41:52 GMT
Server: nginx/1.4.6 (Ubuntu)

Downloading 51.05 MB to "bvlc_googlenet.caffemodel-1"
Done. 51.05 MB in 1653.26288s (31.62 kB/s)
```

The output could be more user friendly in this format:

```
Done. 51.05 MB in 27m 33.26s 
```

tajbojc:
@isidentical it seems that this has been already solved:

**https://github.com/httpie/httpie/blob/master/httpie/output/ui/rich_progress.py#L48-L51**

Should this still be open?

isidentical:
That is correct, thanks for noticing (my changes were isolated, and I didnt thought of this issue but it is nice that it also got solved through)

PR #709, opened 2018-09-08
Make download time more human friendly - Issue #699 fix and Remove exit status check before download - Issue #711 fix

KiranBaktha:


httpie/downloads.py
```diff
@@ -180,6 +181,26 @@ def get_unique_filename(filename, exists=os.path.exists):
...
+    else:
+        time_split = time_string.split(':')
+        time_split[2] = float(time_split[2])
+        if int(time_split[0]):  # More than 1 hour but not a day
+            return "{0[0]}h {0[1]}m {0[2]:.2f}s".format(time_split)
+        elif int(time_split[1]):  # More than 1 minute but not an hour
+            return "{0[1]}m {0[2]:.2f}s".format(time_split)
+        return "{0[2]:.2f}s".format(time_split)  # Less than a min
+
+
```
isidentical, line 203: This seems rather complex for processing that information, especially with multiple branches that identify the same thing. Perhaps you can simplify it like this;
```suggestion
from datetime import timedelta

def friendly_time(total_seconds: float) -> str:
    '''
    Takes in time in seconds and returns a more human friendly time.
    '''
    time_string = str(timedelta(seconds=total_seconds))
    days, _, clock = time_string.rpartition(' days, ')
    hours, minutes, seconds = clock.split(':')
    
    output = ''
    for value, suffix in [
        (days, ' days,'),
        (hours, 'h'),
        (minutes, 'm'),
        (f'{float(seconds):0.2f}', 's'),
    ]:
        if value and float(value):
            output += value + suffix + ' '

    return output
```

isidentical: Even with this, I am not sure how I feel about string processing. Perhaps we can use this instead of manually parsing the output of timedelta;
```py
    days, remainder = divmod(total_seconds, 24 * 60 * 60)
    hours, remainder = divmod(remainder, 60 * 60)
    minutes, seconds = divmod(remainder, 60)
```

This way, we don't need to cast it back to `float()` again.

httpie/core.py
```diff
@@ -143,7 +143,7 @@ def program(args, env, log_error):
                 else:
                     raise
 
-        if downloader and exit_status == ExitStatus.OK:
+        if downloader:
```
isidentical, line 146: Please keep pull requests to a single change only. This seems to be related to another issue, so you can create a new PR for that.

isidentical (changes requested):
Generally good, but let's work on the implementation a bit more.

isidentical:

PR #716, opened 2018-09-24
Update client.py

0xflotus:
What do you think about using the ternary operator in this place?


jakubroztocil:
@0xflotus thanks for the PR, @0xflotus. However, IMO the original, more verbose syntax has the advantage of providing a clearer context for the comment explaining the handling of the `else` condition.

Issue #734, opened 2018-11-29
-F changes the original request method after following redirect


deviantony:
I'm using httpie to execute a POST request against my api at `myapi.mydomain.com/endpoint`:

```
http POST http://myapi.mydomain.com/endpoint         
HTTP/1.1 301 Moved Permanently
Content-Length: 17
Content-Type: text/plain; charset=utf-8
Date: Thu, 29 Nov 2018 21:17:20 GMT
Location: https://myapi.mydomain.com:443/endpoint

Moved Permanently
```

Now using follow:

```
http -F POST http://myapi.mydomain.com/endpoint
HTTP/1.1 405 Method Not Allowed
Content-Length: 0
Date: Thu, 29 Nov 2018 21:18:39 GMT
```

Seems that the request method is changed to GET, when using `curl`, the redirect is working fine:

```
curl -L -X POST http://myapi.mydomain.com/endpoint
{"msg":"hello world"}
```

```
http --version
0.9.8
```

jakubroztocil:
@deviantony changing to `GET` is actually the [correct behaviour](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#Permanent_redirections) (or at least the expected default behaviour on today's web).

(If you, for example, submit a form on a website, the most likely behaviour is that the server redirects you to another "thank you" page after processing your request to avoid double-submit on reload, etc.)

curl's behaviour with `-X POST` is different, but it‘s described as having potential "unintended side-effect" in the docs:

> The method string you set with -X, --request will be used for all requests, which if you for example use -L, --location may cause **unintended side-effects when curl doesn't change request method** according to the HTTP 30x response codes - and similar.
> — https://curl.haxx.se/docs/manpage.html#-X

I assume that curl does change the method only when you use `-d ''` **without** `-X POST`. 

So httpie's default behaviour looks legit to me. What I think it is missing, though, is something like `--data301` (similar to curl's [`--post301` & co](https://curl.haxx.se/docs/manpage.html#--post301)) which would explicitly instruct it to reuse the method and re-send the request data on redirect.

ibnesayeed:
As per the specs, in order for the server to instruct the client to repeat the request on a redirected location with the original HTTP method it should return `307` or `308`. I did not test to see if the `httpie` respects these status codes appropriately.

Issue #592, opened 2017-07-09
Why the Cookie header is removed when follow redirects?


acrazing:
I test as follow:

```bash
http -p H --all --follow github.com AnyHeader:name=value
GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
AnyHeader: name=value
Connection: keep-alive
Host: github.com
User-Agent: HTTPie/0.9.9

GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
AnyHeader: name=value # This header is passed
Connection: keep-alive
Host: github.com
User-Agent: HTTPie/0.9.9


http -p H --all --follow github.com Cookie:name=value
GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Cookie: name=value
Host: github.com
User-Agent: HTTPie/0.9.9

GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: github.com
User-Agent: HTTPie/0.9.9

# The last request without cookie field
```

Env:

```bash
uname -mprs
Darwin 16.5.0 x86_64 i386

http --version
0.9.9
```

PR #736, opened 2018-12-09
[WIP] Add --follow-rule, --post301, --post302 and --post303

matthew16550:
To control what happens when following location redirects. (An idea to solve #592 and #734)

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -52,7 +52,7 @@ class HTTPieRequestsSession(requests.Session):
     def __init__(self, args):
         super(HTTPieRequestsSession, self).__init__()
         self.httpie_current_rule = None
-        self.httpie_follow_rules = args.follow_rules_dict
+        self.httpie_follow_rules = args.follow_rule_dict
         self.httpie_orig_cookies = None
         self.httpie_orig_response_status_code = None
         self.max_redirects = args.max_redirects
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -69,18 +69,18 @@ class HTTPieRequestsSession(requests.Session):
 
     def get_redirect_target(self, resp):
         if self.httpie_follow_rules and 'location' in resp.headers:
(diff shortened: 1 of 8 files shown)


httpie/client.py
```diff
@@ -42,28 +43,68 @@ def init_poolmanager(self, *args, **kwargs):
...
-        transport_plugin = cls()
-        requests_session.mount(prefix=transport_plugin.prefix,
-                               adapter=transport_plugin.get_adapter())
-    return requests_session
+REMOVE_BODY = 0
+KEEP_BODY = codes.temporary_redirect
+
+
+class HTTPieRequestsSession(requests.Session):
+
```
matthew16550: This is pretty much all kludge and I hate myself for doing it but see no other way without getting changes accepted into `requests` first.

matthew16550:
@jakubroztocil I've tried changing `--follow-rule` to work as described in the block below.  But I'm not sure there is a sensible way to implement it without a change to `requests` first so there is an opportunity to suggest a few other changes and get rid of all my kludgyness in `client.py`.  What do you think?

    String specifying how to follow a Location redirect (implies --follow):

        CODE[+HOSTNAME]:METHOD[:nodata][:samecookies]

        "nodata":   Do not resend the request data (by default it is resent).
        "samecookies":  Send exactly the same cookies as the original request.

    e.g. "301:POST" means a POST will be sent after receiving a 301.

    You can specify multiple rules, they will be evaluated in order and the
    first matching rule is used. If no match is found then no follow request
    is sent.

    To avoid accidentally sending data to an unexpected host, a rule such as
    "301:POST" will only match when the redirect location has the same
    hostname as the original request. Rules can specify a different hostname to match
    via the optional HOSTNAME part. HOSTNAME "ALL" means the rule will match
    any host.
    
    TODO something about not redirecting from https to http.

PR #739, opened 2018-12-19
Add compressed requests

aleksandr-vin:
This option allows compression of the uploading files and/or data.

Examples:
```
http --form --compress POST https://localhost/upload csv@./very-big.csv

http -x -x POST https://localhost/upload foo=bar

cat /var/log/system.log | http -x POST https://localhost/upload
```


jakubroztocil:
Thanks @aleksandr-vin!

PR #741, opened 2019-01-04
Fix several ResourceWarning: unclosed file

BoboTiG:
Hello,

This is a patch to fix several resource leaks.

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>


jakubroztocil:
Thanks, Mickaël!

PR #743, opened 2019-01-21
Added test coverage measurement to tox

gaborantal:
Hello there, 

I was curious about the project's test coverage and found out that currently there are no way to measure coverage easily. I did some editing in tox config and now it works perfectly. I thought maybe some other people might be happy if they could measure coverage easily. So I submitted this PR, and hope that you will find it useful and merge it.

Best,
Gábor

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,15 @@
+[run]
+branch = True
+source =
+    httpie
+omit = httpie/tests/*
+
+[report]
+exclude_lines =
+    pragma: no cover
+    except ImportError:
+    def __repr__
+    def __str__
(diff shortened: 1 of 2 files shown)


jakubroztocil:
@gaborantal thank you for the PR! The idea is that tests through `tox` are not covered to make them run faster. [`$ make test`](https://github.com/jakubroztocil/httpie/blob/master/Makefile#L35-L38) runs the tests with coverage, though. See [CONTRIBUTING](https://github.com/jakubroztocil/httpie/blob/master/CONTRIBUTING.rst) for more details.

PR #790, opened 2019-07-18
More graceful messages when stdout or stderr are closed

mgsloan:
This is probably a pretty uncommon case, so not sure if the extra code
is worth it.  But hey, I wrote it so here it is!


jakubroztocil:
Thanks @mgsloan! Something similar is already in `master` and it will shipped with v2.0.0.

PR #791, opened 2019-07-18
Allow stdin to be a closed fd

mgsloan:
Before this change, the following invocation would not work

```
$ http http://neverhttps.com <&-
```

The "<&-" at the end closes the stdin fd. Specifically, it would fail with

```
  ...
  File "/home/mgsloan/.local/lib/python3.6/site-packages/httpie/context.py", line 26, in Environment
    stdin_isatty = stdin.isatty()
AttributeError: 'NoneType' object has no attribute 'isatty'
```

This can occur when httpie is being programmatically invoked, and may
as well be supported.

### DIFF
--- a/httpie/context.py
+++ b/httpie/context.py
@@ -23,7 +23,7 @@ class Environment(object):
     is_windows = is_windows
     config_dir = DEFAULT_CONFIG_DIR
     stdin = sys.stdin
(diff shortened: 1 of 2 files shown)


jakubroztocil:
Thank you!

PR #798, opened 2019-08-16
Fix wrong character output when encoding is not correct

ideal:
Many times web server returns Content-Type without charset info, the encoding returned by
requests will be "ISO-8859-1", however in such situations, it's better to use apparent_encoding instead.

Before:
屏幕快照 2019-08-16 下午3 57 50

After:
屏幕快照 2019-08-16 下午3 59 18

### DIFF
--- a/httpie/models.py
+++ b/httpie/models.py
@@ -42,6 +42,10 @@ class HTTPMessage(object):
 class HTTPResponse(HTTPMessage):
     """A :class:`requests.models.Response` wrapper."""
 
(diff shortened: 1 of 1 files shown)


jakubroztocil:
Thanks @ideal. There’s a problem with touching the response’s `apparent_encoding` attribute, though: it causes that the entire response is downloaded immediately, which breaks how HTTPie promises to work (e.g., to only download what is needed, to stream the output without buffering with `--stream`, etc.).

PR #801, opened 2019-08-27
Update homebrew formula for 1.0.3

jeffbyrnes:
This is a bit of a stab in the dark, but it looks like the Homebrew formula is partially generated by `make brew-deps`, which is then meant to be pasted & committed in `extras/httpie.rb`.

I suspect this file is then also meant to be submitted via a PR to https://github.com/Homebrew/homebrew-core/blob/master/Formula/httpie.rb to get that update out into the world.

I’m happy to take care of that, too, but wanted to submit this here first, to be sure I got this right, as there’s no docs around how to manage this release path.


jakubroztocil:
Thanks

PR #804, opened 2019-09-25
Update installation in Windows

cloudmelon:
Tested in Windows to install httpie using command pip install


jakubroztocil:
Thanks @cloudmelon  but it’s actually already mentioned right above the text you added.

PR #819, opened 2019-12-04
Fix make target in CONTRIBUTING.rst

marhel:
The first hurdle to contributing was that the instructions does not seem up to date. The command `make init` gives:

    make: *** No rule to make target `init'.  Stop. 

Whereas `make install` does what the comment says `make init` was supposed to do (Install dev. requirements and HTTPie in editable mode)


jakubroztocil:
Thanks, @marhel!

PR #821, opened 2019-12-04
run rst2pseudoxml.py with shell=true

roger-wernersson:
makes rst2pseudoxml.py work properly on Windows

executes via a shell instead of not working

also requires that the py extension is associated with the python interpreter


jakubroztocil:
Thanks!

Issue #689, opened 2018-07-13
Downloading "Content-Type: text/plain" files will append .txt at the end of filename


Maryse47:
```
http --debug --download https://raw.githubusercontent.com/jakubroztocil/httpie/master/Makefile
HTTPie 0.9.9
Requests 2.19.1
Pygments 2.2.0
Python 3.6.6 (default) 
[GCC 8.1.1]
/usr/bin/python3
Linux 4.18.0

<Environment {
    "colors": 256,
    "config": {
        "__meta__": {
            "about": "HTTPie configuration file",
            "help": "https://httpie.org/docs#config",
            "httpie": "0.9.9"
        },
        "default_options": "[]"
    },
    "config_dir": "/home/user/.httpie",
    "is_windows": false,
    "stderr": "<_io.TextIOWrapper name='<stderr>' mode='w' encoding='UTF-8'>",
    "stderr_isatty": true,
    "stdin": "<_io.TextIOWrapper name='<stdin>' mode='r' encoding='UTF-8'>",
    "stdin_encoding": "UTF-8",
    "stdin_isatty": true,
    "stdout": "<_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>",
    "stdout_encoding": "UTF-8",
    "stdout_isatty": true
}>

>>> requests.request(**{
    "allow_redirects": true,
    "auth": "None",
    "cert": "None",
    "data": {},
    "files": {},
    "headers": {
        "Accept-Encoding": "identity",
        "User-Agent": "HTTPie/0.9.9"
    },
    "method": "get",
    "params": {},
    "proxies": {},
    "stream": true,
    "timeout": 30,
    "url": "https://raw.githubusercontent.com/jakubroztocil/httpie/master/Makefile",
    "verify": true
})

HTTP/1.1 200 OK
Accept-Ranges: bytes
Access-Control-Allow-Origin: *
Cache-Control: max-age=300
Connection: keep-alive
Content-Length: 3257
Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; sandbox
Content-Type: text/plain; charset=utf-8
Date: xxx GMT
ETag: "xxx"
Expires: xxx GMT
Source-Age: 98
Strict-Transport-Security: max-age=31536000
Vary: Authorization,Accept-Encoding
Via: 1.1 varnish
X-Cache: HIT
X-Cache-Hits: 1
X-Content-Type-Options: nosniff
[...]

PR #891, opened 2020-04-05
Add option to trim extension from downloaded file

kikisix:
#689 
I’m not sure that this issue is still relevant, but anyway - I propose such a fix.
Specifying a flag only for TXT would be too wasteful, I see that forcing trimming the extension can be useful in any other case, even when we can guess the extension by the content type.


isidentical:

PR #897, opened 2020-04-20
add Punycode support to URLs

### LINKED ISSUE #835: Punycode support
I wanted to use https://twitter.com/rfreebern/status/1214560971185778693 with httpie, alas,

```
» http 👻:6677

http: error: InvalidURL: Failed to parse: http://👻:6677
```

Should resolve to `http://xn--9q8h:6677`

Zygimantass:
Theoretically, this should close #835, however this might need testing with cases that I have not thought of.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,5 +38,5 @@ Patches and ideas
 * `Edward Yang <https://github.com/honorabrutroll>`_
 * `Aleksandr Vinokurov <https://github.com/aleksandr-vin>`_
 * `Jeff Byrnes <https://github.com/jeffbyrnes>`_
-
+* `Žygimantas Magelinskas <https://github.com/Zygimantass>`_
 
--- a/httpie/cli/argparser.py
+++ b/httpie/cli/argparser.py
@@ -87,6 +87,12 @@ class HTTPieArgumentParser(argparse.ArgumentParser):
 
         if self.has_stdin_data:
             self._body_from_file(self.env.stdin)
+
+        try:
+            self.args.url = self.args.url.encode("idna").decode("utf-8")
+        except UnicodeError:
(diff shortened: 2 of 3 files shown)


jakubroztocil:
@Zygimantass thanks for the PR! I like the solution but this will require a slightly more careful and deeper integration: it should only operate on the hostname and it should also work for redirects (plus a few other subtleties). I will look into that.

Zygimantass:
@jakubroztocil does the Requests library handle redirects or do we?

Zygimantass:
> 
> 
> I don’t think Requests supports that

I don't mean Punycode support, I mean do we handle following redirects or do they?

jakubroztocil:
@Zygimantass  https://github.com/jakubroztocil/httpie/blob/master/httpie/client.py#L89

jakubroztocil:
Requests [internally uses](https://github.com/psf/requests/blob/9ed5db8ed28e816b597dafd328b342ec95466afa/requests/models.py#L347-L354) the https://pypi.org/project/idna/ library:


> <p>Support for the Internationalised Domain Names in Applications
> (IDNA) protocol as specified in <a href="http://tools.ietf.org/html/rfc5891" rel="nofollow">RFC 5891</a>.
> This is the latest version of the protocol and is sometimes referred to as
> “IDNA 2008”.</p>
> <p>This library also provides support for Unicode Technical Standard 46,
> <a href="http://unicode.org/reports/tr46/" rel="nofollow">Unicode IDNA Compatibility Processing</a>.</p>
> <p>This acts as a suitable replacement for the “encodings.idna” module that
> comes with the Python standard library, but only supports the
> old, deprecated IDNA specification (<a href="http://tools.ietf.org/html/rfc3490" rel="nofollow">RFC 3490</a>).</p>


Which sounds like a better tool for the job than the built-in `idna` codec. I need to investigate the IDNA topic more before merging.

PR #920, opened 2020-05-20
Support (part of) the XDG Base Directory Specification

### LINKED ISSUE #145: Follow the XDG base directory specification
Instead of using config at `$HOME/.httpie`, the default should be `${XDG_CONFIG_HOME:-$HOME/.config}/httpie` for unix-based operating systems according to [the XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

As well as keeping the home directory clean, the semantic separation of config, runtime data, cache data and user-local software by the spec is helpful to users and other tools (e.g. backup software).

In terms of migration, I would suggest using the xdg-compilant config location if it exists and falling back to the current default location. This would also negate the need for the `$HTTPIE_CONFIG_DIR` variable, so it could be removed/de-emphasised from the documentation but still checked for at least a few versions for compatibility.

sersorrel:
On Unix-like systems, the configuration file now lives in `$XDG_CONFIG_HOME/httpie/` by default, not `~/.httpie/` (the behaviour on Windows is unchanged). The previous location is still checked, in order to support existing installations.

Searching `$XDG_CONFIG_DIRS` is still not supported.

Fixes #145; supersedes #436.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1691,8 +1691,11 @@ but you can create it manually.
 Config file directory
 ---------------------
 
-The default location of the configuration file is ``~/.httpie/config.json``
-(or ``%APPDATA%\httpie\config.json`` on Windows).
+The default location of the configuration file on most platforms is
+``$XDG_CONFIG_HOME/httpie/config.json`` (defaulting to
+``~/.config/httpie/config.json``). For backwards compatibility, if the directory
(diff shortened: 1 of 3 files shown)


jakubroztocil:
Awesome, thanks!

Issue #853, opened 2020-02-15
Cookies set to expire in HTTP response are not removed from session file


rca:
When using the `--session` option and storing a session in a file, when the response includes a `Set-Cookie` header that sets a cookie to a date in the past, the cookie is not cleared out of the session file as I would expect.  Because of this, for example, cookies that are set to expire remain in the session and are still sent in subsequent requests.

Thank you.

juped:
@iamcathal and @csr will be working on this!

jakubroztocil:
Sweet, @juped. I've just written some suggestions   on another sessions-related issue that will be useful here as well:

https://github.com/jakubroztocil/httpie/issues/662#issuecomment-641023598

PR #929, opened 2020-06-11
Remove expired cookies

tayoogunbiyi:
This pull request should fix [#853 ](https://github.com/jakubroztocil/httpie/issues/853) @kbanc @juped

### DIFF
--- a/httpie/utils.py
+++ b/httpie/utils.py
@@ -1,10 +1,15 @@
 from __future__ import division
+
 import json
 import mimetypes
+import time
 from collections import OrderedDict
+from http.cookiejar import parse_ns_headers
 from pprint import pformat
 
 import requests.auth
+
+from httpie.models import HTTPResponse
 
 
 def load_json_preserve_order(s):
@@ -83,3 +88,27 @@ def get_content_type(filename):
         if encoding:
             content_type = '%s; charset=%s' % (mime, encoding)
         return content_type
+
+
+def get_expired_cookies(response: HTTPResponse) -> list:
+    original = response._orig.raw._original_response
+
+    expired_cookies = []
+    cookie_headers = []
+    curr_timestamp = time.time()
+
+    for header in original.msg._headers:
+        if header[0] == 'Set-Cookie':
+            cookie_headers.append(header[1])
+
+    extracted_cookies = parse_ns_headers(cookie_headers)
+
+    for cookie in extracted_cookies:
+        cookie_name = cookie[0][0]
+        for cookie_key, cookie_value in cookie:
+            if cookie_key == 'expires':
+                is_expired = curr_timestamp > cookie_value
+                if is_expired:
+                    expired_cookies.append(cookie_name)
+
+    return expired_cookies
--- a/httpie/utils.py
+++ b/httpie/utils.py
@@ -9,6 +9,7 @@ from pprint import pformat
 
 import requests.auth
 
+
 def load_json_preserve_order(s):
     return json.loads(s, object_pairs_hook=OrderedDict)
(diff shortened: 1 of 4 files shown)


httpie/utils.py
```diff
@@ -83,3 +86,25 @@ def get_content_type(filename):
...
         return content_type
+
+
+def get_expired_cookies(response):
+    original = response.raw._original_response
+    expired_cookies = []
+    cookie_headers = []
+    curr_timestamp = time.time()
+
+    for header_name, content in original.msg._headers:
```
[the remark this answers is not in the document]
kbanc: We couldn't find a parser for the `Set-Cookie` header string that is returned in the `response.headers`. While `parse_ns_headers` and `cookie.SimpleCookie` works ok if there's only one cookie in the `response.headers`, they doesn't parse multiple cookies correctly.

tests/test_sessions.py
```diff
@@ -186,3 +189,69 @@ def test_download_in_session(self, httpbin):
                  httpbin.url + '/get', env=self.env())
         finally:
             os.chdir(cwd)
+
+
+class TestExpiredCookie(SessionTestBase):
```
jakubroztocil, line 194: Plural "cookies" would be better. Also, it doesn’t need to inherit from `SessionTestBase` (which would make sense to be renamed to `MultiStepSessionTestBase`).

tests/test_sessions.py
```diff
@@ -186,3 +189,69 @@ def test_download_in_session(self, httpbin):
...
+
+    @pytest.mark.parametrize(
+        argnames=['initial_cookie', 'expired_cookie'],
+        argvalues=[
+            ({'id': {'value': 123}}, 'id'),
+            ({'id': {'value': 123}}, 'token')
+        ]
+    )
+    def test_removes_expired_cookies_from_session_obj(self, initial_cookie, expired_cookie, httpbin):
+        self.start_session(httpbin)
```
jakubroztocil, line 204: Calling `start_session` should be unnecessary here.

tests/test_sessions.py
```diff
@@ -186,3 +189,69 @@ def test_download_in_session(self, httpbin):
...
+        ([('Set-Cookie', 'hello=world; Path=/; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly'),
+            ('Set-Cookie', 'pea=pod; Path=/ab; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly'),
+            ('Connection', 'keep-alive')],
+            None,
+            [{'name': 'hello', 'path': '/'}, {'name': 'pea', 'path': '/ab'}]),
+        ([('Set-Cookie', 'hello=world; Path=/; Expires=Fri, 12 Jun 2020 12:28:55 GMT; HttpOnly'),
+            ('Connection', 'keep-alive')],
+            datetime(2020, 6, 11).timestamp(),
+            [])
+    ]
```
jakubroztocil, line 254: With this formatting it‘s a bit hard to understand the structure of the fixture data. Could you use indents?

tests/test_sessions.py
tayoogunbiyi: plural "cookies" definitely is better. However, the inheritance from SessionTestBase seems necessary due to the fact that the test involves creating a session file

tayoogunbiyi: It may involve duplicating the logic in the `SessionTestBase.teardown_method(self,method)` function, as I'd have to clean up the session file created during the test

jakubroztocil:
Thanks! 🍰

PR #931, opened 2020-06-13
Cookie redirect update fix 

### LINKED ISSUE #662: Cookies not correctly updated within a --follow redirect chain
Hello,

when site creates new session for user, the PHPSESSID is not overrided but both the old and the new one is sent. Please notice Cookie header in last request. I'm using httpie version 0.9.8.

```
» http --verbose --session=/tmp/yamaha.json --form --follow POST https://www.yamaha-extranet.com/login/index email=foo@bar.baz password=bar submitform=Submit
POST /login/index HTTP/1.1
Cookie: PHPSESSID=mdslhb7u0giujsaf8itq2gm2p0
Host: www.yamaha-extranet.com
User-Agent: HTTPie/0.9.8

email=****&submitform=Submit

HTTP/1.1 302 Found
Location: /
Set-Cookie: PHPSESSID=nb7qnhkfjsdtpe8gtj797koaq7; path=/; domain=www.yamaha-extranet.com; secure
X-Powered-By: PHP/5.5.38



GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Cookie: PHPSESSID=mdslhb7u0giujsaf8itq2gm2p0; PHPSESSID=nb7qnhkfjsdtpe8gtj797koaq7
Host: www.yamaha-extranet.com
User-Agent: HTTPie/0.9.8
```

Session file contains only one PHPSESSID though.
```
{
    "__meta__": {
        "about": "HTTPie session file",
        "help": "https://httpie.org/docs#sessions",
        "httpie": "0.9.8"
    },
    "auth": {
        "password": null,
        "type": null,
        "username": null
    },
    "cookies": {
        "PHPSESSID": {
            "expires": null,
            "path": "/",
            "secure": true,
            "value": "nb7qnhkfjsdtpe8gtj797koaq7"
        }
    },
    "headers": {}
}
```

asifmallik:
Fixes #662 

The crux of the bug is that `httpie` was not saving the domain of the cookie. As a result, when a session was loaded from a file and a request was sent to a server that sets a new cookie of the same name, its value would not be overwritten.

I have not written tests for this. I think there are two relevant things that need to be tested:

1. Testing that cookies from different domains do not overwrite each other
2. Testing that the headers are correct

I am not exactly sure how to test 1 as it seems to me that everything is happening locally with httpbin so direction would be appreciated. As for 2, my original idea was to write a parser for the headers in `BaseCLIResponse` and to make assertions on the request header in `test_session_update`. 

Is this a good idea?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 * Added ``--ciphers`` to allow configuring OpenSSL ciphers (`#870`_).
 * Added support for ``$XDG_CONFIG_HOME`` (`#920`_).
 * Added support for custom content types for uploaded files (`#668`_).
-
+* Fixed cookie not updated during redirect by saving domain for cookies (`#662`).
 
 `2.1.0`_ (2020-04-18)
 ---------------------
--- a/httpie/sessions.py
+++ b/httpie/sessions.py
@@ -103,7 +103,7 @@ class Session(BaseConfigDict):
     @cookies.setter
     def cookies(self, jar: RequestsCookieJar):
         # <https://docs.python.org/2/library/cookielib.html#cookie-objects>
-        stored_attrs = ['value', 'path', 'secure', 'expires']
+        stored_attrs = ['value', 'domain', 'path', 'secure', 'expires']
         self['cookies'] = {}
         for cookie in jar:
             self['cookies'][cookie.name] = {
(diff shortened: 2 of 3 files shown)


asifmallik:
Instead of parsing the entirety of the request header, to test for the bug, I just asserted that "Cookie: hello=world2" is in the CLI response as if the bug where to be present it would be "Cookie: hello=world; hello=world2" since this is the kind of assertions I saw in other tests

jakubroztocil:
@asifmallik what about https://github.com/jakubroztocil/httpie/issues/662#issuecomment-642722138 ?

asifmallik:
@jakubroztocil The issue there again, as pointed out in [psf/requests#5489](https://github.com/psf/requests/issues/5489), is that the cookie was not properly attributed to the correct domain using, I believe, cookiejar. When session is used, I could not reproduce this issue.

Issue #824, opened 2019-12-08
Cookies from original request cannot be combined with response cookies in session file


strindberg:
Consider the following request:

```bash
https --session=/tmp/c-session "https://localhost:8721/customer/business/1" Cookie:sAuth=foo6
```

If the server sets cookies `XSRF-TOKEN` and `JSESSIONID`, the session file will look like this:

```json
{
    "__meta__": {
        "about": "HTTPie session file",
        "help": "https://httpie.org/doc#sessions",
        "httpie": "1.0.3"
    },
    "auth": {
        "password": null,
        "type": null,
        "username": null
    },
    "cookies": {
        "JSESSIONID": {
            "expires": null,
            "path": "/",
            "secure": true,
            "value": "091642DF767443D96E72C6FDEE561428"
        },
        "XSRF-TOKEN": {
            "expires": null,
            "path": "/",
            "secure": true,
            "value": "af6eb371-ce07-4583-bdce-efbfa09728f9"
        }
    },
    "headers": {
        "Cookie": "sAuth=foo6"
    }
}
```

When the request is repeated with the same session file (but without the sAuth given on the command line), the result is that only the cookie sAuth is sent, not the cookies `JSESSIONID` and `XSRF-TOKEN`:

```bash
https --verbose --session=/tmp/c-session "https://localhost:8721/apis/customer/business/1"
```

Request:

```http
GET /apis/customer/business/1 HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Cookie: saamAuth=foo6
Host: localhost:8721
User-Agent: HTTPie/1.0.3
```

I would have expected to have all three cookies set in the request.

kbanc:
@jakubroztocil Can I look into this?

jakubroztocil:
@kbanc sure thing! (As always, I’d start by writing a test case reproducing the behavior and clarifying the problem.)

isidentical:
If I am not missing anything, this seem to be fixed by @kbanc on #932. If the issue is still persisting, please let us know by creating another ticket.

PR #932, opened 2020-06-15
Combine cookies from original request and session file

kbanc:
This pull request should address #824. Cookies that are set through a request in the CLI are added to the session cookies. @gmelodie

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1760,7 +1760,6 @@ To set a cookie within a Session there are three options:
 3. Manually set cookie parameters in the json file of the session
 
 .. code-block:: json
-    :emphasize-lines: 12-19
     
     {
         "__meta__": {
--- a/README.rst
+++ b/README.rst
@@ -530,7 +530,6 @@ Simple example:
     $ http PUT httpbin.org/put name=John email=john@example.org
 
 .. code-block:: http
-
     PUT / HTTP/1.1
     Accept: application/json, */*;q=0.5
     Accept-Encoding: gzip, deflate
@@ -1738,6 +1737,61 @@ exchange after it has been created, specify the session name via
 
     # But it is not updated:
     $ http --session-read-only=./ro-session.json httpbin.org/headers Custom-Header:new-value
+
+Cookie Storage Behaviour
+------------------------
+
+**TL;DR:** Cookie storage priority: Server response > Command line request > Session file 
+
+To set a cookie within a Session there are three options: 
+
+1. Get a `Set-Cookie` header in a response from a server
+
+.. code-block:: bash
+
+    $ http --session=./session.json httpbin.org/cookie/set?foo=bar
+
+2. Set the cookie name and value through the command line as seen in `cookies`_
+
+.. code-block:: bash
+
+    $ http --session=./session.json httpbin.org/headers Cookie:foo=bar
+
+3. Manually set cookie parameters in the json file of the session
+
+.. code-block:: json
+    :emphasize-lines: 12-19
+    
+    {
+        "__meta__": {
+        "about": "HTTPie session file",
+        "help": "https://httpie.org/doc#sessions",
+        "httpie": "2.2.0-dev"
+        },
+        "auth": {
+            "password": null,
+            "type": null,
+            "username": null
+        },
+        "cookies": {
+            "foo": {
+                "expires": null,
+                "path": "/",
+                "secure": false,
+                "value": "bar"
+                }
+        }
+    }
+
+Cookies will be set in the session file with the priority specified above. For example, a cookie
+set through the command line will overwrite a cookie of the same name stored 
+in the session file. If the server returns a `Set-Cookie` header with a
+cookie of the same name, the returned cookie will overwrite the preexisting cookie.
+
+Expired cookies are never stored. If a cookie in a session file expires, it will be removed before
+sending a new request. If the server expires an existing cookie, it will also be removed from the
+session file. 
+
 
 Config
 ======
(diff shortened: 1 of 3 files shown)


httpie/sessions.py
```diff
@@ -79,6 +79,13 @@ def update_headers(self, request_headers: RequestHeadersDict):
             if name == 'User-Agent' and value.startswith('HTTPie/'):
                 continue
 
+            if name == 'Cookie':
```
jakubroztocil, line 82: There’s no guarantee on the capitalization (i.e., you probably wanna `.lower()` the name before comparing it).

httpie/sessions.py
```diff
@@ -79,6 +79,13 @@ def update_headers(self, request_headers: RequestHeadersDict):
...
+                for cookie in value.split(';'):
+                    name, value = cookie.split("=")
+                    self['cookies'][name.strip()] = {'value': value.strip()}
+                del request_headers['Cookie']
+                continue
+
```
jakubroztocil, line 88: Isn’t there an existing implementation of the `Cookie` header value parsing? I’d imagine `requests` or the standard library having one.

jakubroztocil: The parser should also gracefully handle situations when the header contains invalid value (e.g., `Cookie: hello` or `Cookie: ;;;===`)—people often intentionally send invalid requests with HTTPie to test the server’s behavior.

jakubroztocil: If you end up using a custom implementation, then I’d probably move it to a function (and add some unittests for it).

httpie/sessions.py
jakubroztocil: Actually, maybe there’s is some implied guaranteed since the it seems to work for the `User-Agent` check above :) But since header names are generally considered case-insensitive, it would be good to always normalize before comparing header names.

httpie/sessions.py
```diff
@@ -79,6 +81,12 @@ def update_headers(self, request_headers: RequestHeadersDict):
             if name == 'User-Agent' and value.startswith('HTTPie/'):
```
jakubroztocil, line 81: This is pre=existing code, but please change it to use the same header name normalization like you use for `Cookie` bellow for consistency.

gmelodie:
Hey @jakubroztocil we worked on some refactoring today. We added what you mentioned on 
bfea90f and then we did some refactoring on the tests on 5b98e4a. If you could tell us which you like best I'd be great!

httpie/sessions.py
```diff
@@ -76,7 +78,13 @@ def update_headers(self, request_headers: RequestHeadersDict):
...
 
             value = value.decode('utf8')
-            if name == 'User-Agent' and value.startswith('HTTPie/'):
+            if name.lower() == 'user-agent' and value.startswith('HTTPie/'):
+                continue
+
+            if name.lower() == 'cookie':
+                for cookie_name, morsel in SimpleCookie(value).items():
+                    self['cookies'][cookie_name] = {'value': morsel.value}
+                del request_headers['Cookie']
```
jakubroztocil, line 87: Here it should use the value from `name`, not `'Cookie'`.

httpie/sessions.py
```diff
@@ -79,6 +81,12 @@ def update_headers(self, request_headers: RequestHeadersDict):
             if name == 'User-Agent' and value.startswith('HTTPie/'):
                 continue
 
+            if name.lower() == 'cookie':
+                for cookie_name, morsel in SimpleCookie(value).items():
+                    self['cookies'][cookie_name] = {'value': morsel.value}
```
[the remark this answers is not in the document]
kbanc: We added a new test, `test_cookie_storage_priority` in TestSessions to define and document the priority rules! Is this sufficient, or should there also be documentation in the docs?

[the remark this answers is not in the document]
kbanc: After our call this morning, @gmelodie and I discussed the possible behaviours of cookies expiring in sessions and which cookies should take precedent.

Our understanding was regardless of the expiry behaviour, the storage priority should still be server set > command line set > session file. For that reason, we didn't write further related tests. Here's a couple of example cases illustrating why we didn't think further tests were necessary:

-  Case 1: Server expires a cookie. If a server expires a cookie, it will be removed from the session file regardless of how it was set (`test_sessions.py` already has test covering this)
- Case 2: Cookies in session file expire. Cookies stored in a session file that expire will never be sent to the server and will be removed when the request is made
- Case 3: User sets an expired cookie in the command line. We haven't been able to find a way for a user to set an expired cookie in the command line. However, if one is set, we would assume the intention would be to remove the cookie from the session. As this cookie would replace the cookie in the session file, and expired cookies are cleared from the session file when the request is sent, this behaviour is satisfied.

README.rst
```diff
@@ -1739,6 +1738,60 @@ exchange after it has been created, specify the session name via
...
+2. Set the cookie name and value through the command line as seen in `cookies`_
+
+.. code-block:: bash
+
+    $ http --session=./session.json httpbin.org/headers Cookie:foo=bar
+
+3. Manually set cookie parameters in the json file of the session
+
+.. code-block:: json
+
```
kbanc, line 1763: We tried to emphasize lines 12-18 in this code block using :emphasize-lines:, but we weren't able to get it to render properly.

httpie/sessions.py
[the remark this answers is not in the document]
jakubroztocil: > Our understanding was regardless of the expiry behaviour, the storage priority should still be server set > command line set > session file. 

Agree 👍🏻

jakubroztocil:
Awesome work, thanks @kbanc and @gmelodie! 🍰 🚀

Issue #852, opened 2020-02-14
.netrc not honored if auth-type is used


pszlazak:
Once `--auth-type` switch is used, `.netrc` is not honored. Authentication details must be provided via `--auth` switch.

Details:
```
# http --debug --auth-type=basic example.org
HTTPie 0.9.8
Requests 2.19.1
Pygments 2.2.0
Python 2.7.13 (default, Sep 26 2018, 18:42:22)
[GCC 6.3.0 20170516]
/usr/bin/python
Linux 4.9.0-9-amd64

<Environment {
    "colors": 8,
    "config": {
        "__meta__": {
            "about": "u'HTTPie configuration file'",
            "help": "u'https://httpie.org/docs#config'",
            "httpie": "u'0.9.8'"
        },
        "default_options": "[]"
    },
    "config_dir": "/root/.httpie",
    "is_windows": false,
    "stderr": "<open file '<stderr>', mode 'w' at 0x7fd6490091e0>",
    "stderr_isatty": true,
    "stdin": "<open file '<stdin>', mode 'r' at 0x7fd6490090c0>",
    "stdin_encoding": "UTF-8",
    "stdin_isatty": true,
    "stdout": "<open file '<stdout>', mode 'w' at 0x7fd649009150>",
    "stdout_encoding": "UTF-8",
    "stdout_isatty": true
}>
usage: http [--json] [--form] [--pretty {all,colors,format,none}]
            [--style STYLE] [--print WHAT] [--headers] [--body] [--verbose]
            [--all] [--history-print WHAT] [--stream] [--output FILE]
            [--download] [--continue]
            [--session SESSION_NAME_OR_PATH | --session-read-only SESSION_NAME_OR_PATH]
            [--auth USER[:PASS]] [--auth-type {basic,digest}]
            [--proxy PROTOCOL:PROXY_URL] [--follow]
            [--max-redirects MAX_REDIRECTS] [--timeout SECONDS]
            [--check-status] [--verify VERIFY]
            [--ssl {ssl2.3,tls1,tls1.1,tls1.2}] [--cert CERT]
            [--cert-key CERT_KEY] [--ignore-stdin] [--help] [--version]
            [--traceback] [--default-scheme DEFAULT_SCHEME] [--debug]
            [METHOD] URL [REQUEST_ITEM [REQUEST_ITEM ...]]
http: error: --auth required
```

nb2838:
I'd like to tackle this if that's ok!

[...]

PR #934, opened 2020-06-16
Honor .Netrc when --auth-type is used

nb2838:
I added code so that httpie honors .netrc when --auth-type is used. This was done by explicitly handling the .netrc rather than implicitly leaving it to the requests library. Maybe it would be good to in the future disallow reading the .netrc if the file has too many permissions as is done with some applications (this is also the default behaviour of the python library.) I also added a test to check that the behaviour does work. 

#852


jakubroztocil:
@nb2838 Going through this realized there was a bunch of netrc-related issues. So I made a bigger commit addressing them all all. I cherry-picked the test you wrote. Thanks for contributing!

PR #938, opened 2020-06-23
Add test to test auth plugin reused in session

### LINKED ISSUE #937: Add test for auth_plugin and session behaviour
I noticed that there doesn't seem to be code coverage for the use of auth plugins (`--auth-type`) in a session. 

Creating an issue to add a test in `test_sessions.py` to cover it.

kbanc:
Hoping to resolve #937 with this PR. 

Looking for feedback to verify that the first test (`test_auth_type_reused_in_session`) is a valid test as it is currently failing.

### DIFF
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -216,6 +216,9 @@ class TestSession(SessionTestBase):
 
     def test_auth_type_reused_in_session(self, httpbin):
         self.start_session(httpbin)
+        session_path = self.config_dir / 'test-session.json'
+
+        header = 'Custom dXNlcjpwYXNzd29yZA'
 
         class Plugin(AuthPlugin):
             auth_type = 'test-reused'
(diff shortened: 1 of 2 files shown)


tests/test_sessions.py
```diff
@@ -216,33 +216,38 @@ def test_download_in_session(self, httpbin):
...
+        header = 'Custom dXNlcjpwYXNzd29yZA'
 
         class Plugin(AuthPlugin):
             auth_type = 'test-reused'
             auth_require = False
 
             def get_auth(self, username=None, password=None):
                 assert self.raw_auth == 'user:password'
-                return basic_auth(header='Custom dXNlcjpwYXNzd29yZA==')
+                return basic_auth(header='{}=='.format(header))
```
jakubroztocil, line 238: instead of `.format()`, you can use an f-strings: `f'{header}=='`

jakubroztocil:
Thanks, @kbanc! 🍰 And sorry for the wait.

PR #945, opened 2020-06-27
Added quiet flag

### LINKED ISSUE #926: Add 'quiet' or 'silent' option for using httpie in scripts
Simple flag to avoid unnecessary output when using inside scripts.

nb2838:
Added a quiet flag, tests to verify it works and documentation. 

The code works by writing to `os.devnull` as requested. 

Right now the only problem is that I let the python garbage collector close the file instead of me explicitly closing it. I tried putting a statement to close the file at different parts in the code (specifically in `core.py`) but when using `--download` and `--quiet` at the same time the program breaks and I get an error specifying that I am trying to write to a closed file.

Nevertheless, as it stands right now the code works properly. Fixes #926


httpie/context.py
```diff
@@ -34,6 +37,8 @@ class Environment:
     stdout_encoding: str = None
     stderr: IO = sys.stderr
     stderr_isatty: bool = stderr.isatty()
+    devnull: IO = None
+    devnull_isatty: bool = False
```
jakubroztocil, line 41: `devnull_isatty` is never used

tests/test_downloads.py
```diff
@@ -189,3 +189,11 @@ def test_download_with_redirect_original_url_used_for_filename(self, httpbin):
             assert os.listdir('.') == [expected_filename]
         finally:
             os.chdir(orig_cwd)
+
+    def test_download_quietflag(self, httpbin_both, httpbin):
```
jakubroztocil, line 193: 'quietflag'  does not sound right

tests/test_downloads.py
```diff
@@ -189,3 +189,11 @@ def test_download_with_redirect_original_url_used_for_filename(self, httpbin):
...
+        robots_txt = '/robots.txt'
+        body = urlopen(httpbin + robots_txt).read().decode()
```
jakubroztocil, line 195: please use `requests`

tests/test_downloads.py
```diff
@@ -189,3 +189,11 @@ def test_download_with_redirect_original_url_used_for_filename(self, httpbin):
...
+        env = MockEnvironment(stdin_isatty=True, stdout_isatty=False)
+        r = http('--quiet', '--download', httpbin_both.url + robots_txt, env=env)
+        assert r.stderr == ''
+        assert body == r
```
jakubroztocil, line 199: you should also test that both `env.std(out|err)` get replaced with `env.devnull`

httpie/context.py
```diff
@@ -34,6 +37,8 @@ class Environment:
     stdout_encoding: str = None
     stderr: IO = sys.stderr
     stderr_isatty: bool = stderr.isatty()
+    devnull: IO = None
```
jakubroztocil, line 40: `devnull` should be cached property that opens the file on first access

nb2838:
Thank you! will make the requested changes. The unclosing problem doesn't manifest itself as the code is right now. However, if I open devnull on first access then it does manifest itself when I use the --download flag for example. That is why the PR I submitted didn't open devnull on first-access. I will write more tests that show this behavior.

jakubroztocil:
> The unclosing problem doesn't manifest itself as the code is right now. 

Is it really a problem then?

jakubroztocil:
I guess it could be done through [`__del__()`](https://docs.python.org/3/reference/datamodel.html#object.__del__) on `Environment`, something like we do here:

https://github.com/jakubroztocil/httpie/blob/0541490dda1c5c9eaa87a4a542fe55e41c990742/tests/utils.py#L80-L86

But then Python is probably already about to close the file on its own:

> If you don’t explicitly close a file, **Python’s garbage collector will eventually destroy the object and close the open file** for you, but the file may stay open for a while. Another risk is that different Python implementations will do this clean-up at different times.
> — https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files

nb2838:
> > The unclosing problem doesn't manifest itself as the code is right now.
> 
> Is it really a problem then?

I think it would be a problem if I do modify the code as you suggest here 
> `devnull` should be cached property that opens the file on first access
what you suggest later might work though so I will look into that. I also thought that just letting the garbage collector do its job would work but as I said, for some reason it fails when we use --download and --quiet at the same time and I open the file inside of the environment object. I will try to investigate and see how to fix it. 

The other option would be just to not open devnull inside the Environment object but I agree that seems ugly and bad practice so it would better be avoided.

jakubroztocil:
> I think it would be a problem if I do modify the code as you suggest here

why?

nb2838:
> > I think it would be a problem if I do modify the code as you suggest here
> 
> why?
Actually, I am sorry. I misread what you suggested. I think it does work.

nb2838:
If I understand properly what you suggest is using the function in functools cached_property. I just checked and you were right it does work! Nevertheless, this is only supported since python 3.8 which would require updating the requirements a bit as they are now set for python3.6 or above. Is this ok?

jakubroztocil:
I’m not referring to the specific Python 3.8 implementation, but to the general concept. It’s too small of a reason to bump the required version.

jakubroztocil:
@nb2838 thanks for the changes!

The behavior when a file is being saved with `--quiet` is under-defined or incorrect:

```bash
$ http --quiet --download httpbin.org/get
```
```bash
$ http --quiet --download --ouptput=foo.json httpbin.org/get
```

```bash
$ http --quiet --ouptput=./foo.json httpbin.org/get
```

All these result in an empty file being created.

The quietness should probably only apply to `stdout` and `stderr`, and not to output files. (Even if it applied to output files as well, creating an empty file doesn’t feel right.)

What do you think?

nb2838:
@jakubroztocil Yes, I definitely agree!
It should definitely only apply to stdout and stderr. I hadn't noticed this behaviour before with the quite flag. I will try to fix it and maybe look into the issue you mentioned if necessary.  Also thank you for all your feedback and suggestions.

jakubroztocil:
@nb2838 Perfect. A good place to start would be to fully define the correct behavior and write the docs (README + CLI help) and also the tests first. For things like these, the implementation then tends to be the easiest part.

nb2838:
@jakubroztocil I was just running the commands that you told me to  you had issues with and I seem to only have the issue you mentioned with
`` $ http --quiet --output=./foo.json httpbin.org/get``
the other two seem to be fine. What do you think might be causing this issue? I tested it from the terminal so maybe if we tested it in different ways we might get different behavior. Here is my terminal session 
``` zsh
(httpie) ➜  check-download git:(quiet) ✗ ls
(httpie) ➜  check-download git:(quiet) ✗ http --quiet --download httpbin.org/get
(httpie) ➜  check-download git:(quiet) ✗ ls
get.json
(httpie) ➜  check-download git:(quiet) ✗ cat get.json
{
  "args": {},
  "headers": {
    "Accept": "*/*",
    "Accept-Encoding": "identity",
    "Authorization": "Basic aHR0cGllOnRlc3Q=",
    "Host": "httpbin.org",
    "User-Agent": "HTTPie/2.3.0-dev",
    "X-Amzn-Trace-Id": "Root=1-5f15a48c-56066a1a015fd4e90413de7b"
  },
  "origin": "186.28.38.162",
  "url": "http://httpbin.org/get"
}
(httpie) ➜  check-download git:(quiet) ✗ http --quiet --download --output=foo.json httpbin.org/get
(httpie) ➜  check-download git:(quiet) ✗ ls
foo.json get.json
(httpie) ➜  check-download git:(quiet) ✗ cat foo.json
{
  "args": {},
  "headers": {
    "Accept": "*/*",
    "Accept-Encoding": "identity",
    "Authorization": "Basic aHR0cGllOnRlc3Q=",
    "Host": "httpbin.org",
    "User-Agent": "HTTPie/2.3.0-dev",
    "X-Amzn-Trace-Id": "Root=1-5f15a4d0-875f70b4f0b3f7a53c538974"
  },
  "origin": "186.28.38.162",
  "url": "http://httpbin.org/get"
}
(httpie) ➜  check-download git:(quiet) ✗ http --quiet --output=./foo.json httpbin.org/get
(httpie) ➜  check-download git:(quiet) ✗ cat ./foo.json
(httpie) ➜  check-download git:(quiet) ✗ ls
foo.json get.json
```

nb2838:
I added the changes you requested and it works. I would like to include some tests to check that it works but I couldn't find any tests in the codebase related to checking the content of downloaded file. I was thinking of maybe just creating a temporary file in the test, checking that the contents that are written are correct, and then deleting it. I was not sure if you thought this was the best idea though so I wanted to first ask and get your approval or suggestions on how this could be done.

jakubroztocil:
@nb2838 thanks for the changes. As for the tests, you start with something like this, that additionally checks the contents:

https://github.com/httpie/httpie/blob/bece3c77bb51ecc55dcc4008375dc29ccd91575c/tests/test_downloads.py#L181-L191

jakubroztocil:
@nb2838 your commits have been [merged](https://github.com/httpie/httpie/compare/a4a1e8...2c7f24), thanks! 🥳 

I polished the code / docs a bit before merging it, and also made it so errors are still printed to `stderr`. 

You can see my last changes here: https://github.com/httpie/httpie/commit/1aa1366f9957601344302f4de194ad4e7bf683ae

PR #947, opened 2020-07-01
Add workflow to test with pyOpenSSL active

### LINKED ISSUE #939: Add a build environment where `pyOpenSSL` is active
Under very specific circumstances, `requests` will use [`pyOpenSSL`](https://www.pyopenssl.org/):

> 2.24.0 (2020-06-17)
> `pyOpenSSL` TLS implementation is now only used if Python either doesn't have an `ssl` module or doesn't support SNI. Previously `pyOpenSSL` was unconditionally used if available. This applies even if `pyOpenSSL` is installed via the `requests[security]` 
> — https://github.com/psf/requests/blob/master/HISTORY.md#2240-2020-06-17

We have at least two places in the codebase where this is significant.

#### 1. different error types are used

https://github.com/jakubroztocil/httpie/blob/49e71d252f54871a6bc49cb1cba103d385a543b8/tests/test_ssl.py#L10-L22

#### 2. difference  messages are used (despite the comment incorrectly talking about differences in architectures): 

https://github.com/jakubroztocil/httpie/blob/49e71d252f54871a6bc49cb1cba103d385a543b8/tests/test_ssl.py#L136-L141

---

It would be great to update [.github/workflows/build.yml](https://github.com/jakubroztocil/httpie/blob/master/.github/workflows/build.yml) so that there’s also an environment where `pyopenssl` is guaranteed to be loaded by `requests`. That would catch errors like [this one](https://marc.info/?l=openbsd-ports&m=159251948515635&w=2).

gmelodie:
Monkeypatches the `pyOpenSSL` module using [`urllib3.contrib.pyopenssl.inject_into_urllib3()`](https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/pyopenssl.py#L20-L30). Testing is done using an environment variable (`HTTPIE_TEST_WITH_OPENSSL`) which, if enabled (`HTTPIE_TEST_WITH_OPENSSL="1"`), will trigger a pytest fixture to do the monkeypatching.

Fixes #939.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-      - run: python -m pip install --upgrade pip
+      - run: pip install --upgrade pip setuptools wheel
       - run: pip install --upgrade --editable .
       - run: python setup.py test
         env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
         python-version: [3.6, 3.7, 3.8]
+        pyopenssl: ["0", "1"]
         exclude:
           - os: windows-latest
             python-version: 3.8
@@ -32,6 +33,8 @@ jobs:
       - uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-      - run: pip install --upgrade pip setuptools wheel
+      - run: python -m pip install --upgrade pip
       - run: pip install --upgrade --editable .
       - run: python setup.py test
+        env:
+          HTTPIE_TEST_WITH_PYOPENSSL: ${{ matrix.pyopenssl }}
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 from pytest_httpbin import certs
 
@@ -22,3 +23,19 @@ def httpbin_secure_untrusted(monkeypatch, httpbin_secure):
     """
     monkeypatch.delenv('REQUESTS_CA_BUNDLE')
     return httpbin_secure
+
+
+@pytest.fixture(autouse=True, scope='session')
+def pyopenssl_inject():
+    """
+    Injects `pyOpenSSL` module to make sure `requests`
+    will use it.
+
+    <https://github.com/psf/requests/pull/5443#issuecomment-645740394>
+    """
+
+    if os.getenv('HTTPIE_TEST_WITH_PYOPENSSL', default='0') != '1':
+        return
+
+    import urllib3.contrib.pyopenssl
+    urllib3.contrib.pyopenssl.inject_into_urllib3()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,9 +33,11 @@ def pyopenssl_inject():
 
     <https://github.com/psf/requests/pull/5443#issuecomment-645740394>
     """
-
-    if os.getenv('HTTPIE_TEST_WITH_PYOPENSSL', default='0') != '1':
-        return
-
     import urllib3.contrib.pyopenssl
     urllib3.contrib.pyopenssl.inject_into_urllib3()
+
+    if os.getenv('HTTPIE_TEST_WITH_PYOPENSSL', default='0') == '0':
+        urllib3.contrib.pyopenssl.extract_from_urllib3()
+
+    yield
+    urllib3.contrib.pyopenssl.extract_from_urllib3()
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -32,6 +33,8 @@ jobs:
       - uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
+        env:
+          HTTPIE_TEST_WITH_OPENSSL: ${{ matrix.pyopenssl }}
```
jakubroztocil, line 37: I’d rename the variable to `HTTPIE_TEST_WITH_PYOPENSSL`.

tests/conftest.py
```diff
@@ -22,3 +23,20 @@ def httpbin_secure_untrusted(monkeypatch, httpbin_secure):
...
+def pyopenssl_inject():
+    """
+    Injects `pyOpenSSL` module to make sure `requests`
+    will use it.
+
+    <https://github.com/psf/requests/pull/5443#issuecomment-645740394>
+    """
+
+    if os.getenv('HTTPIE_TEST_WITH_OPENSSL', default='0') == '1':
+        try:
```
jakubroztocil, line 38: Here I’d rather try to explicitly import PyOpenSSL directly and show an error that it needs to be installed. The injecting part itself should not be inside a try…catch.

tests/conftest.py
```diff
@@ -22,3 +23,20 @@ def httpbin_secure_untrusted(monkeypatch, httpbin_secure):
...
+            import urllib3.contrib.pyopenssl
+            urllib3.contrib.pyopenssl.inject_into_urllib3()
+        except ImportError:
+            raise ImportError("Unable to import pyOpenSSL")
```
jakubroztocil, line 42: `'`

gmelodie:
Wrote a test (`test_pyopenssl`) to check if `HTTPIE_TEST_WITH_PYOPENSSL` is set and `pyOpenSSL` is being used ([the same way that `urllib3` tests it](https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/pyopenssl.py#L124-L125))

jakubroztocil:
Hi @gmelodie, I see the tests are failing. Are you still working on this?

gmelodie:
@jakubroztocil Yes! I'll get down to that today. Last couple of days have been a bit messy.

jakubroztocil:
No worries, I was just curious.

gmelodie:
@jakubroztocil the tests seem to be working (they catch an error on windows). I'm not sure whether fixing this issue is in scope of this PR. If you think so, just give me a heads up and I'll try and work on that later today :)

tests/test_ssl.py
```diff
@@ -138,3 +140,17 @@ def test_ciphers_none_can_be_selected(httpbin_secure):
...
+
+    assert env_with_pyopenssl == '0' or env_with_pyopenssl == '1'
+
+    if env_with_pyopenssl == '0':
+        assert urllib3.util.ssl_.IS_PYOPENSSL is False
+        assert urllib3.util.IS_PYOPENSSL is False
+
+    elif env_with_pyopenssl == '1':
+        assert urllib3.util.ssl_.IS_PYOPENSSL is True
+        assert urllib3.util.IS_PYOPENSSL is True
```
BoboTiG, line 156:
```suggestion
    using_pyopenssl = os.getenv('HTTPIE_TEST_WITH_PYOPENSSL', '0')
    if using_pyopenssl == '0':
        assert not urllib3.util.ssl_.IS_PYOPENSSL
        assert not urllib3.util.IS_PYOPENSSL
    else:
        assert urllib3.util.ssl_.IS_PYOPENSSL
        assert urllib3.util.IS_PYOPENSSL
```

tests/conftest.py
```diff
@@ -22,3 +23,21 @@ def httpbin_secure_untrusted(monkeypatch, httpbin_secure):
...
+    <https://github.com/psf/requests/pull/5443#issuecomment-645740394>
+    """
+    import urllib3.contrib.pyopenssl
+    urllib3.contrib.pyopenssl.inject_into_urllib3()
+
+    if os.getenv('HTTPIE_TEST_WITH_PYOPENSSL', default='0') == '0':
+        urllib3.contrib.pyopenssl.extract_from_urllib3()
+
+    yield
+    urllib3.contrib.pyopenssl.extract_from_urllib3()
```
BoboTiG, line 43:
```suggestion
    if os.getenv('HTTPIE_TEST_WITH_PYOPENSSL', '0') == '1':
        try:
            import urllib3.contrib.pyopenssl
            urllib3.contrib.pyopenssl.inject_into_urllib3()
        except ImportError as exc:
            pytest.fail(f"Could not import PyOpenSSL: {exc!r}")

    yield
```

BoboTiG (changes requested):
@gmelodie could you rebase your PR on the master branch? I also left a review, that would be great if you could take it into account :)

Issue #948, opened 2020-07-01
Tests fail when running openssl with security level > 1


matthewhughes934:
Hi,

I noticed an environment specific test-failure when I was poking around this repo.

PS. the setup process of just running `make` to get everything ready for development is really nice, thanks for that.

**Description**

The following tests fail when running openssl with security level > 1:

`tests/test_ssl.py::TestClientCert::test_cert_and_key`
`tests/test_ssl.py::TestClientCert::test_cert_pem`

The failures end with `[SSL: CA_MD_TOO_WEAK] ca md too weak (_ssl.c:3824)` (though I can share the entire output if needed)

**Cause**

The cause appears to be `tests/client_certs/client.crt` being signed with `sha1WithRSAEncryption`, from the [openssl docs](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_get_security_level.html):

> SHA1 algorithm is very widely used in certificates and will be rejected at levels higher than 1 because it only offers 80 bits of security.

I've further confirmed this by reducing the security level of openssl on my system (Debian 10, which defaults to security level 2) and observing the tests passing.

**Resolution**

Possible resolutions:

* Do nothing, since this is likely to impact only a small portion of people wishing to run these tests. Maybe document the failure to notify others
* Re-create test certificate and sign with a more secure algorithm
* Skip these tests if the security level is lower than expected (though at a glance there's not a nice interface for determining this from the Python's `ssl` lib)

If it's decided that this is worth actioning I'd be happy to attempt whatever action is decided.

PR #952, opened 2020-07-04
Quieten ssl tests

matthewhughes934:
I noticed some warnings and failures in `tests/test_ssl.py` when running the test suite.

There were two failures because Debian has stricter than default OpenSSL settings (security level 2, minimum protocol TLSv1.2, see also Issue #948 ).


matthewhughes934:
I need to fix failures for Python < 3.7

jakubroztocil:
Thank you, @matthewhughes934! 🎉

PR #957, opened 2020-07-23
Add shell completion support via argcomplete

scop:
Could replace the extras/ completions with this I think. And if argcomplete dependency isn't desired, I suppose it could be made optional.

### DIFF
--- a/httpie/cli/definition.py
+++ b/httpie/cli/definition.py
@@ -5,6 +5,8 @@ CLI arguments definition.
 from argparse import (FileType, OPTIONAL, SUPPRESS, ZERO_OR_MORE)
 from textwrap import dedent, wrap
 
+from argcomplete.completers import ChoicesCompleter, FilesCompleter
+
 from .. import __doc__, __version__
 from .argparser import HTTPieArgumentParser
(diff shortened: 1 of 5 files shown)


httpie/cli/definition.py
```diff
@@ -68,7 +70,8 @@
         $ http example.org hello=world   # => POST
 
     '''
-)
+).completer = ChoicesCompleter(
+    ('GET', 'HEAD', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'))
```
LuckyDenis, line 74: Good evening. Is there a reason why it's not in the list: CONNECT, TRACE?

scop: CONNECT is proxy stuff only, not sure why one would use it directly with httpie (no idea whether it'd actually work). And TRACE is such a rarity that I don't think it deserves to be there. The thinking is to offer the most commonly used choices as completions. But if there's disagreement and CONNECT and TRACE actually work, sure they could be added.

TECH90HS (approved):
Noce

scop:
Rebased on current `master`.

PR #968, opened 2020-09-19
Delete .editorconfig

JilderCalderonGT:


jakubroztocil:
Thank you for the PR, but we do like our `.editorconfig`.

PR #991, opened 2020-11-12
Support directory input for --output

### LINKED ISSUE #789: FR: Support using directory for "--output"
This would be useful when used with download mode and continue.

Since continue forces us to pass the output file it means the file name would need to be static, but in some cases (Scripts) we might need to still get the file name from the server (or guess it) and continue if a partial file already exists.

In cases where the server provided Content-Disposition what I do now is get that then use it for the "-o" filename, but when this is not an option things get more complicated so having this would help.

akshgpt7:
Resolves #789.

The following implementation would allow users to provide directories as well as files for the `--output` option's value.
It works this way for the following cases:
- With `http -do <directory> <url>`, it creates the download file inside the given directory using the same file naming logic as before.
- With `http -dco <directory> <url>`, it searches for the file to continue inside the directory provided, using the name that would have been created during download.
- With `http -o <directory> <url>`, it creates the output file inside the given directory using the same naming logic as downloaded file name.

An error is raised in case a non-existing directory is passed.
**The existing behavior of passing a file as `--output` is still the same.**

### DIFF
--- a/httpie/cli/definition.py
+++ b/httpie/cli/definition.py
@@ -415,7 +415,7 @@ output_options.add_argument(
 )
 output_options.add_argument(
     '--output', '-o',
-    type=FileType('a+b'),
+    type=str,
     dest='output_file',
     metavar='FILE',
     help='''
--- a/httpie/cli/definition.py
+++ b/httpie/cli/definition.py
@@ -2,7 +2,7 @@
 CLI arguments definition.
 
 """
-from argparse import (FileType, OPTIONAL, SUPPRESS, ZERO_OR_MORE)
+from argparse import (OPTIONAL, SUPPRESS, ZERO_OR_MORE)
 from textwrap import dedent, wrap
 
 from httpie import __doc__, __version__
--- a/httpie/cli/definition.py
+++ b/httpie/cli/definition.py
@@ -416,12 +416,12 @@ output_options.add_argument(
 output_options.add_argument(
     '--output', '-o',
     type=str,
-    dest='output_file',
+    dest='output_dest',
     metavar='FILE',
     help='''
-    Save output to FILE instead of stdout. If --download is also set, then only
-    the response body is saved to FILE. Other parts of the HTTP exchange are
-    printed to stderr.
+    Save output to a FILE or create it inside a DIRECTORY instead of stdout.
+    If --download is also set, then only the response body is saved to FILE.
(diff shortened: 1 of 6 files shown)


gmelodie (approved):
LGTM

httpie/cli/definition.py
```diff
@@ -415,13 +415,13 @@
 )
 output_options.add_argument(
     '--output', '-o',
-    type=FileType('a+b'),
-    dest='output_file',
+    type=str,
```
flosincapite, line 418: Did you look in to making a new type for this? Did that not turn out to be feasible?

flosincapite (commented):
Sorry for the late review--I didn't see this for some reason.

Just one comment on the `FileType` thing, as we discussed.

httpie/cli/definition.py
akshgpt7: I think that checking for a valid string is all that might be needed here. That's because if the user enters any given string, it will create a download file with that given name. So I think the only check we need to make is for a valid string. (Let me know if you have differing viewpoints)

gracewgao: Like Aksh said, the current --output allows the creation of file with the string that's passed in, so not much needs to be changed now that we've just changed the argument type to string.

flosincapite:
Cool, looks like you just have an error to address to get the tests passing:

```
  File "/home/runner/work/httpie/httpie/tests/test_output.py", line 306, in test_json_formatting_options
    r = http(
  File "/home/runner/work/httpie/httpie/tests/utils.py", line 244, in http
    exit_status = main(args=complete_args, **kwargs)
  File "/home/runner/work/httpie/httpie/httpie/core.py", line 62, in main
    parsed_args = parser.parse_args(
  File "/home/runner/work/httpie/httpie/httpie/cli/argparser.py", line 88, in parse_args
    self._setup_standard_streams()
  File "/home/runner/work/httpie/httpie/httpie/cli/argparser.py", line 156, in _setup_standard_streams
    if not os.path.isdir(self.output_dest):
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/genericpath.py", line 42, in isdir
    st = os.stat(s)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

AntonNiklasson:
Looks like it's been stale for a while. I would love to see this merged. 👏

PR #1000, opened 2020-12-02
Fix documentation on file upload

paretje:
As documented later on in "File upload forms", the correct syntax to set
the mimetype of the upload is `field@file;type=filetype`


jakubroztocil:
Good catch, thanks @paretje! 🥧

Issue #996, opened 2020-11-30
Command breaks pip installation on Windows: pip install --upgrade pip setuptools


xcodz-dot:
directly using pip breaks the installation

the Usage given in README.md:
```bash
pip install --upgrade pip setuptools
```

the usage which must be given:
```bash
python -m pip install --upgrade pip setuptools
```

PR #1005, opened 2020-12-21
Update README.rst for fixxing the issue #996

xcodz-dot:
fixes: #996


jakubroztocil:
Thanks, @xcodz-dot! 🥧

Issue #965, opened 2020-09-17
Incomplete responses do not cause an error


dbrgn:
It seems that httpie does not validate the number of received bytes.

Here are some explanations: https://blog.petrzemek.net/2018/04/22/on-incomplete-http-reads-and-the-requests-library-in-python/

With this demo script:

```python
#!/usr/bin/env python3
#
# An HTTP server that returns fewer bytes in the body of the response than
# stated in the Content-Length header.
#

import socketserver

class MyTCPHandler(socketserver.BaseRequestHandler):
    def handle(self):
        data = self.request.recv(1024)
        print(data.decode())
        self.request.sendall(
            b'HTTP/1.1 200 OK\r\n'
            b'Content-Length: 10\r\n'
            b'\r\n'
            b'123456'
        )

class MyTCPSever(socketserver.TCPServer):
    allow_reuse_address = True

with MyTCPSever(('localhost', 8080), MyTCPHandler) as server:
    server.serve_forever()
```

...httpie happily returns the incomplete response without raising an error.

```
$ http GET :8080/
HTTP/1.1 200 OK
Content-Length: 10

123456

$ echo $?
0
```

On the other hand, curl raises an error:

```
$ curl http://127.0.0.1:8080/
curl: (18) transfer closed with 4 bytes remaining to read
123456$
```

More details can be found in the blogpost. It seems that requests 3 will contain an option to force validation of incomplete responses, but it's not yet released. There are also workarounds for the current behavior.

jakubroztocil:
Thanks for the report and for including the test case.

We perform the length validation when `--download`ing:

```bash
$ http --download :8080
HTTP/1.1 200 OK
Content-Length: 10

Downloading 10.00 B to "index"
http: error: Incomplete download: size=10; downloaded=6

Done. 6.00 B in 0.00053s (11.12 kB/s)
$ echo $?
1
```

And I agree it would make sense to apply the validation for regular invocations as well. Especially with `--check-status`.

PR #1009, opened 2021-01-03
Displaying error for incomplete responses

nathanielCherian:
In response to issue #965. A simple addition that compares the content-length header and response length. Apologies if there is a better solution as I am still exploring this project.

Continuing the test from the issue... 
```
$ http GET :8080/
http: error: Incomplete read: size=10; read=6

HTTP/1.1 200 OK
Content-Length: 10

123456
```

### DIFF
--- a/httpie/core.py
+++ b/httpie/core.py
@@ -220,7 +220,6 @@ def program(
                             level='warning'
                         )
 
-
                 if not downloader:
                     # repeating check for check_status specific
                     if(message.headers.get("Content-Length") and int(message.headers.get("Content-Length")) != len(message.content or "")):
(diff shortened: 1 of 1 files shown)


jakubroztocil:
Thanks, @nathanielCherian! It looks good in principle (even though I don’t see any tests). But let me think through how this fits into some bigger changes that are planned regarding the output machinery.

nathanielCherian:
Sounds good! I'll create some tests if you choose to include it.

jakubroztocil:
@nathanielCherian I’ve opened a new issue for this feature request (#1030). Implementation-wise, we’ll need to find a way how to keep track of received bytes without touching `response.content` because of streaming.

PR #1014, opened 2021-01-09
Progress bar toggle

### LINKED ISSUE #889: Allow disabling progress bar for logging in Download mode
In `Download mode`, httpie adds a progress bar which is printed along with the response headers. Both are logged to `stderr` preventing the progress bar from being skipped while logging response to a logfile using say output redirection.

I'd suggest either an argument to disable progress bar, `--no-progress-bar` or print the progress bar to `stdout` so that only `stderr` can be logged to files and `stdout` can be send to ~~hell~~ `/dev/null`

For now I forced to do this instead :(
image

PerfectlyInternal:
Resolves #889 

Adds a new CLI argument --no-progress-bar, which disables the progress bar. Works exactly as @arjunv requested.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,5 +38,5 @@ Patches and ideas
 * `Edward Yang <https://github.com/honorabrutroll>`_
 * `Aleksandr Vinokurov <https://github.com/aleksandr-vin>`_
 * `Jeff Byrnes <https://github.com/jeffbyrnes>`_
-
+* `PerfectlyInternal <https://github.com/perfectlyinternal>`_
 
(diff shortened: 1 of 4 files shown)


jakubroztocil:
Thanks, @PerfectlyInternal! It looks good in principle (even though I don’t see any tests). But let me think through how this fits into some bigger changes that are planned regarding the output machinery.

PerfectlyInternal:
I figured tests weren't necessary for a minor change to the way output works, as it is entirely cosmetic and has no effect on the actual function of the command. If tests are necessary, I'd be happy to implement them ASAP.

Issue #1008, opened 2020-12-31
http: error: UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 1617: character maps to <undefined> when re-direction output to disk


peterkronenberg:
On Windows 10, I'm using httpie to retrieve data. It works fine when displaying on the screen, but as soon as I pipe the output to disk, I get http: error: UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 8024: character maps to <undefined>

I tried to attach the file I'm having the problem with, but unfortunately, it 100Mb

Here is the complete debug output:

```
C:\Users\PeterKronenberg\Desktop>http --debug --print=Bb --pretty=all --session=C:\Users\PETERK~1\AppData\Local\Temp/access.json http://localhost:8080/nexus/extract/path  filePath=c:/TestFiles/100Mb.pdf > foo
HTTPie 2.3.0
Requests 2.25.1
Pygments 2.7.3
Python 3.9.1 (tags/v3.9.1:1e5d33e, Dec  7 2020, 17:08:21) [MSC v.1927 64 bit (AMD64)]
c:\python\python39\python.exe
Windows 10

<Environment {'colors': 256,
 'config': {'default_options': []},
 'config_dir': WindowsPath('C:/Users/PeterKronenberg/AppData/Roaming/httpie'),
 'devnull': <property object at 0x000001112C1815E0>,
 'is_windows': True,
 'log_error': <function Environment.log_error at 0x000001112C1AC1F0>,
 'program_name': 'http',
 'stderr': <colorama.ansitowin32.StreamWrapper object at 0x000001112C176F10>,
 'stderr_isatty': True,
 'stdin': <_io.TextIOWrapper name='<stdin>' mode='r' encoding='utf-8'>,
 'stdin_encoding': 'utf-8',
 'stdin_isatty': True,
 'stdout': <colorama.ansitowin32.StreamWrapper object at 0x000001112C176BE0>,
 'stdout_encoding': 'cp1252',
 'stdout_isatty': False}>

>>> requests.request(**{'auth': None,
 'data': '{"filePath": "c:/TestFiles/100Mb.pdf"}',
[...]

PR #1025, opened 2021-01-21
Document workaround for UnicodeDecodeError on Windows: redirect problems with `--pretty` (refs #1008)

Almad:
#1008

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1613,6 +1613,17 @@ by adding the following to your ``~/.bash_profile``:
         http --pretty=all --print=hb "$@" | less -R;
     }
 
+If you enforce `--pretty=all` with redirected output and you get `UnicodeDecodeError`,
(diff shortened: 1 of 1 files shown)


jakubroztocil:
Thanks, @Almad. It’s bit more nuanced:

1. That issue relates specifically to colors, which can be enabled with both `--format=all` and `--format=colors` 
2. It is specific to Windows. 
3. It may have to do more with how colors are represented on Windows than with an encoding mismatch.

I’d find out what actually gets written to the file when colors are applied. And then see if it even makes sense to allow non-terminal coloured output on Windows: does `cat ./output` correctly reproduce the coloured output?  Perhaps the best solution will be simply disabling colors on Windows unless the output is a terminal.

Almad:
It makes sense to not pollute the docs—this should be sufficiently googleable for people looking for workaround. Closing until we do a real fix, which is to research coloring on Windows and potentially disable color output when on Windows and redirecting to a file.

Issue #998, opened 2020-12-01
Will not expire/delete cookie from session when Set-Cookie only sets Max-Age=0 without Expires


segevfiner:
1. Run the following flask app (Python 3+, Flask 1.1.0):
```python
from flask import Flask


app = Flask(__name__)


@app.route('/set')
def set_():
    return "Set cookie", {'Set-Cookie': 'test=test; Path=/; Max-Age=3600'}


@app.route('/expire')
def expire():
    return "Expired cookie", {'Set-Cookie': 'test=test; Path=/; Max-Age=0'}
```
2. `http --session=./session.json http://localhost:5000/set` - The cookie will be set
3. `http --session=./session.json http://localhost:5000/expire` - The cookie will not be unset from the session file, check out its contents.

Flask does auto-set `Expires` when using `set_cookie` with only `max_age`, but not all web frameworks do that, and I don't think it is required by the standard.

LuckyDenis:
@jakubroztocil, Good afternoon! 
I found out what the problem is. We lose information about expired `cookies` in this place, since there is no check for the `max-age` parameter.

What is in the variable `cookies = [{'path': '/', 'max-age': '0', 'version': '0', 'name': 'test'}]`

https://github.com/httpie/httpie/blob/cf78a12e46ced68003263f4f43cee8fac720aa2c/httpie/utils.py#L112-L119


The reason why we don't have the `expires` variable set in the cookie is that when receiving a response, the `requests` library doesn't handle `max-age=0`. 

```
>>> import requests
>>> r = requests.get('http://localhost:5000/expire')
>>> r.cookies
<RequestsCookieJar[]>
>>> r.headers
{'Set-Cookie': 'test=test; Path=/; Max-Age=0', 'Content-Type': 'text/html; charset=utf-8', 'Content-Length': '14', 'Server': 'Werkzeug/1.0.1 Python/3.8.7', 'Date': 'Thu, 04 Feb 2021 12:55:55 GMT'}

```

It seems to me that this is a bug in `requests`, since we lose the state of the cookies. I suggest that at the moment create a issue in `requests` and insert a solution in the likeness of this with the addition of a link to the issue in `requests`.

```
cookies = [
        # The first attr name is the cookie name.
        dict(attrs[1:], name=attrs[0][0])
        for attrs in attr_sets
    ]

    # issue ...
    for cookie in cookies:
        max_age = cookie.get('max-age')
        if max_age and max_age.isdigit():
[...]

PR #1029, opened 2021-02-05
Fix (#998): Will not expire/delete cookie from session when Set-Cookie only sets Max-Age=0 without Expires

LuckyDenis:
Summary:
The reason why the `cookie` state information is lost is that when a response is received, the query library does not process `max-age=0`. It was decided to report the problem to the `requests` community and implement a temporary solution.

Done:
- [x] Fix issue #998
- [x] Add test for issue
- [x] Run `make test-all`
- [x] Make https://github.com/psf/requests/issues/5743

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,5 +38,6 @@ Patches and ideas
 * `Edward Yang <https://github.com/honorabrutroll>`_
 * `Aleksandr Vinokurov <https://github.com/aleksandr-vin>`_
 * `Jeff Byrnes <https://github.com/jeffbyrnes>`_
+* `Denis Belavin <https://github.com/LuckyDenis>`_
 
 
--- a/httpie/utils.py
+++ b/httpie/utils.py
@@ -109,6 +109,15 @@ def get_expired_cookies(
         for attrs in attr_sets
     ]
 
+    # HACK/FIXME: https://github.com/psf/requests/issues/5743
+    for cookie in cookies:
+        if 'expires' in cookie:
+            continue
+
+        max_age = cookie.get('max-age')
+        if max_age and max_age.isdigit():
+            cookie['expires'] = now + float(max_age)
+
     return [
         {
             'name': cookie['name'],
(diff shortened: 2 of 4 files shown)


LuckyDenis:
I was unable to repeat this error on my macos. Can anyone try to reproduce it, and throw off the full error's log? 😓

httpie/utils.py
```diff
@@ -109,6 +109,15 @@ def is_expired(expires: Optional[float]) -> bool:
...
 
+    # HACK/FIXME: https://github.com/psf/requests/issues/5743
+    for cookie in cookies:
+        if 'expires' in cookie:
+            continue
+
+        max_age = cookie.get('max-age')
+        if max_age and max_age.isdigit():
+            cookie['expires'] = now + float(max_age)
+
```
jakubroztocil, line 120: This would be better as a new named function with a docstring for context & clarity.

jakubroztocil (changes requested):

jakubroztocil:
Thanks, @LuckyDenis 🥧

LuckyDenis:
@jakubroztocil, Thank you, I hope to participate in your wonderful project again ^_^

AUTHORS.rst
```diff
@@ -38,5 +38,6 @@ Patches and ideas
 * `Edward Yang <https://github.com/honorabrutroll>`_
 * `Aleksandr Vinokurov <https://github.com/aleksandr-vin>`_
 * `Jeff Byrnes <https://github.com/jeffbyrnes>`_
+* `Denis Belavin <https://github.com/LuckyDenis>`_
```
george79d, line 41: []()

george79d: Thx

httpie/utils.py
```diff
@@ -109,6 +109,15 @@ def is_expired(expires: Optional[float]) -> bool:
...
     return [
```
george79d, line 112:
```suggestion
     return [
```

PR #1032, opened 2021-02-13
Correctly handle single-byte Content-Range

blyxxyz:
HTTPie fails if it continues a download with a single byte left. For example:
```
$ http -b -d example.org
Downloading 1.23 kB to "index.html"
Done. 1.23 kB in 0.00043s (2.81 MB/s)
$ python3 -c 'import sys; sys.stdout.write(open("index.html").read()[:-1])' > index-short.html
$ http -b -d -c -o index-short.html example.org

http: error: ContentRangeError: Invalid Content-Range returned: 'bytes 1255-1255/1256'
```
The range is inclusive on both ends.


jakubroztocil:
@blyxxyz good catch. Thanks for the PR and test! 🥧

PR #1040, opened 2021-02-24
Replace typography quotes (fixes #1039)

### LINKED ISSUE #1039: pip install error
The **setup.cfg** in the main folder of this project has a pair of unusual character which will result the following Error output
I add a print statement in dist.py it turns out to out put the file name:**setup.cfg**
图片
So i run cat setup.cfg which prints this 
图片
I delete single quotes and everything went normal.

Almad:


jakubroztocil:
Fixes #1039

PR #1042, opened 2021-02-27
Fail gracefully if multiple request data files are supplied

blyxxyz:
HTTPie currently crashes if you give it multiple request data files:
```
$ http : @/dev/null @/dev/null
[...]
    fn, fd, ct = self.args.files['']
ValueError: not enough values to unpack (expected 3, got 2)
```
This replaces that by an error message.


jakubroztocil:
Thanks, @blyxxyz!

PR #1054, opened 2021-04-30
Replace usage of mock with unittest.mock

BoboTiG:
Since Python 3, the `mock` dependency is no more required as it is already part of the `unittest` module.

I am not sure it needs a changelog entry, let me know :)


jakubroztocil:
Thanks, @BoboTiG! 🥧

PR #1058, opened 2021-05-03
Prefer usage of pytest rather than py.test

BoboTiG:
`py.test` was chosen over `pytest` but it is not planned for removal yet [1].
Anyway, it is a good thing to ensure we are using the correct Python version
with the right `pytest` installed, so using `python -m pytest` is recommended.

[1] https://github.com/pytest-dev/pytest/issues/1629


jakubroztocil:
Cheers, @BoboTiG!

PR #1061, opened 2021-05-07
Add --data-raw argument to POST data

### LINKED ISSUE #534: Request to add `-d, --data` option for raw body like curl
The request is simple, just to add an option to pass raw data like `curl` does:

    http :/api/user -d 'MyRawData...'

I know that in mostly cases if you are sending JSON or form data, it can be achieved with the _"request items"_, like:

    http :/api/hey say=Hello to=me …

And it will converted to the proper format depending of the content type, that is awesome! And if you have something that is not a JSON or form data to send, you can do something like:

    echo 'MyRawData...' | http :/api/hey

But this is impractical, the main idea of HTTPie is _cURL-like tool for humans_ , and this case is far of that principle, in fact, `curl` is more practical than HTTPie for the previous example. Adding more than one command and pipe them with ugly characters like `|` `<` just because a simple option is missing doesn't sound _human-friendly_.

What's wrong with add the `-d` option to `http`?

BoboTiG:
A first draft for `curl --data-raw` support. It would close #534.

Notes:
1. `--data-raw` has been chosen because it is the one from `curl --data-raw` we are emulating here.
2. I wanted to use a short argument version, like `-d`, but it is already used for `--download, -d`. I initially used `-D`, and can restore it if you prefer.
   A possible solution would be to rename `--download, -d` to `--download, -dl`, but it may not be possible due to possible breakings everywhere :)
   I thought we could go with `-D` and then it would be kind of logic to use `--data-binary, -Db` later.
3. No need to implement `curl --data-urlencode 'name=I am Daniel'` as it is implicitely handled.
4. That PR does not handle `curl --data 'birthyear=1905&press=OK'`.
5. That PR does not handle `curl --data-binary ''`.

---

## Single `--data-raw`

Reference:
```shell
$ curl --data 'birthyear=1905' -XPOST pie.dev/post
```
```json
{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {
    "birthyear": "1905"
  }, 
  "headers": {
    "Content-Length": "14", 
    "Content-Type": "application/x-www-form-urlencoded", 
    "User-Agent": "curl/7.64.1"
  }, 
  "json": null, 
  "origin": "...", 
  "url": "http://pie.dev/post"
}
```

New argument `--data-raw`:
```shell
$ http --data-raw 'birthyear=1905' POST pie.dev/post
```
```json
{
    "args": {},
    "data": "",
    "files": {},
    "form": {
        "birthyear": "1905"
    },
    "headers": {
        "Content-Length": "14",
        "Content-Type": "application/x-www-form-urlencoded; charset=utf-8",
        "User-Agent": "HTTPie/2.5.0-dev"
    },
    "json": null,
    "origin": "...",
    "url": "http://pie.dev/post"
}
```

## Multiple `--data-raw`

Reference:
```shell
$ curl --data 'birthyear=1905' --data 'press=OK' -XPOST pie.dev/post
```
```json
{
  "args": {}, 
  "data": "", 
  "files": {}, 
  "form": {
    "birthyear": "1905", 
    "press": "OK"
  }, 
  "headers": {
    "Content-Length": "23", 
    "Content-Type": "application/x-www-form-urlencoded", 
    "User-Agent": "curl/7.64.1"
  }, 
  "json": null, 
  "origin": "...", 
  "url": "http://pie.dev/post"
}
```

New argument `--data-raw`:
```shell
$ http --data-raw 'birthyear=1905' --data-raw 'press=OK' POST pie.dev/post
```
```json
    {
        "args": {},
        "data": "",
        "files": {},
        "form": {
            "birthyear": "1905",
            "press": "OK"
        },
        "headers": {
            "Content-Length": "23",
            "Content-Type": "application/x-www-form-urlencoded; charset=utf-8",
            "User-Agent": "HTTPie/2.5.0-dev"
        },
        "json": null,
        "origin": "...",
        "url": "http://pie.dev/post"
    }
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -227,6 +227,13 @@ Submitting `forms`_:
     $ http -f POST pie.dev/post hello=World
 
 
+Submitting `forms`_, alternative way:
+
+.. code-block:: bash
+
+    $ http --data-raw 'hello=World' POST pie.dev/post
+
+
 See the request that is being sent using one of the `output options`_:
 
 .. code-block:: bash
@@ -481,8 +488,9 @@ their type is distinguished only by the separator used:
 +------------------------------+---------------------------------------------------+
 | Data Fields                  | Request data fields to be serialized as a JSON    |
 | ``field=value``,             | object (default), to be form-encoded              |
-| ``field=@file.txt``          | (with ``--form, -f``), or to be serialized as     |
-|                              | ``multipart/form-data`` (with ``--multipart``).   |
+| ``field=@file.txt``          | (with ``--form, -f`` or ``--data-raw``), or to    |
+|                              | be serialized as ``multipart/form-data`` (with    |
+|                              | ``--multipart``).                                 |
 +------------------------------+---------------------------------------------------+
 | Raw JSON fields              | Useful when sending JSON and one or               |
 | ``field:=json``,             | more fields need to be a ``Boolean``, ``Number``, |
@@ -673,6 +681,21 @@ Regular forms
     Content-Type: application/x-www-form-urlencoded; charset=utf-8
 
     name=John+Smith
+
+
+Regular forms, alternative way
+------------------------------
+
+.. code-block:: bash
+
+    $ http --data-raw 'name=John Smith' --data-raw 'press=OK' POST pie.dev/post
+
+.. code-block:: http
+
+    POST /post HTTP/1.1
+    Content-Type: application/x-www-form-urlencoded; charset=utf-8
+
+    name=John+Smith&press=OK
 
 
 File upload forms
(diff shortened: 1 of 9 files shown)


BoboTiG:
It will now be possible to do such things:
```shell
$ http --data-raw 'a=b' POST pie.dev/post more=data
```
It is a workaround for that foberidden use:
```shell
echo 'data' | http POST example.org more=data   # This is invalid
```

BoboTiG:
Closing the PR as it is not answering the orignal need. Will open a new one with the approppriate behavior.

PR #1062, opened 2021-05-08
Add --raw to allow specifying the raw request body without extra processing

BoboTiG:
Addressing #534.

In short, the new argument allows to use that synxtax:
```shell
$ http --raw '{"name": "John"}' pie.dev/post
```

Instead of:
```shell
$ echo '{"name": "John"}' | http pie.dev/post
$ http pie.dev/post <<<'{"name": "John"}'
```


BoboTiG:
There is a case not handled:
```shell
$ echo -n 'χρυσαφὶ' | http -r 'data' POST httpbin.org/post
```
In that case, STDIN is silently ignored.

jakubroztocil:
Thanks, @BoboTiG!

> There is a case not handled:
> 
> ```shell
> $ echo -n 'χρυσαφὶ' | http -r 'data' POST httpbin.org/post
> ```
> 
> In that case, `STDIN` is silently ignored.

I think it would be more consistent to throw an error like we do when you combine `STDIN` with data fields (`echo a | http httpbin.org/post a=a`).

BoboTiG:
The documentation is now less disparate. WDYT @elenalape and @jakubroztocil ?

elenalape (changes requested):
Really excited about this PR being shipped! 🚢 Have a look at the proposed doc changes and let me know what you think.

README.rst
```diff
@@ -1456,6 +1463,19 @@ verbatim contents of that XML file with ``Content-Type: application/xml``:
 File uploads are always streamed to avoid memory issues with large files.
 
 
+Request data from the CLI
```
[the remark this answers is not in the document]
BoboTiG: Yes, why not. I put that section last as per Jakub's comment (via email). If we go from the easier to the more complex, it makes sens then.

README.rst
```diff
@@ -1353,8 +1353,16 @@ which you don’t care about. The response headers are downloaded always,
 even if they are not part of the output
 
 
+Specifying the raw body
+=======================
+
+Several solutions are available to allow passing the raw body of the request.
+Each one has its own pros and cons following the use case, pick the one you
+are more confortable with.
```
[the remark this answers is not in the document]
BoboTiG: Let's remove the pros and cons part :)

BoboTiG:
Thanks for the review @elenalape.
I applied your suggestions, do you have a better phrasing to propose maybe?

README.rst
```diff
@@ -1353,8 +1353,29 @@ which you don’t care about. The response headers are downloaded always,
...
+
+An alternative for situations when ``stdin`` is not convenient to use
+is to specify the raw request body via the ``--request, -r`` argument.
+
+.. code-block:: bash
+
+    $ http --request '{"name": "John"}' pie.dev/post
+
+The data is passed as-is to the request without processing — it is not
+converted into `JSON`_ or `forms`_ data.
```
jakubroztocil, line 1374: I’d use the following order: `stdin`, `--request`, `@file`.

Here’s my suggestion for the raw request intro as well as for the `--request` section (which should be moved after `stdin`):

```suggestion
Specifying raw request body
============================

In addition to crafting structured `JSON`_ and `forms`_ requests with the `request items`_ syntax, you can provide a raw request body that will be sent without further processing. These two approaches specifying request data (i.e., structured and raw) cannot be combined.

There’re three methods for passing raw request data: piping via ``stdin``, ``--request='data'``, and ``@/file/path``.

Request data via ``--request``
------------------------------

In a situation when piping data via ``stdin`` is not convenient (for example, when generating API docs examples), you can specify the raw request body via the ``--request, -r`` option.

.. code-block:: bash

    $ http --request 'Hello, world!' pie.dev/post

.. code-block:: bash

    $ http --request '{"name": "John"}' pie.dev/post
    
```

httpie/cli/definition.py
```diff
@@ -185,6 +185,13 @@
 
     '''
 )
+content_type.add_argument(
+    '--request', '-r',
+    help='''
+    Specify the raw request body without extra processing whatsoever.
```
jakubroztocil, line 191:
```suggestion
    This option allows you to pass raw request data without extra processing 
    (as opposed to the structured request items syntax):
     
        $ http --request='data' pie.dev/post
        
    You can achieve the same by piping the data via stdin:
    
        $ echo data | http pie.dev/post
        
    Or have HTTPie load the raw data from a file:
    
        $ http pie.dev/post @data.txt
    
```

CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 `2.5.0-dev`_ (unreleased)
 -------------------------
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
+* Added ``--request, -r`` to allow specifying the raw request body without extra processing whatsoever. (`#534`_)
```
jakubroztocil, line 13:
```suggestion
* Added ``--request, -r`` to allow specifying the raw request body without extra processing as an alternative to `stdin`. (`#534`_)
```

jakubroztocil (changes requested):
I’ve added some suggestions and expanded the docs a bit. Otherwise, it looks good. Now let’s see what the final option name should be, and then the PR can be finalized and merged.

README.rst
BoboTiG: Way better ^^

CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 `2.5.0-dev`_ (unreleased)
 -------------------------
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
+* Added ``--request, -r`` to allow specifying the raw request body without extra processing as an alternative to `stdin`. (`#534`_)
```
BoboTiG, line 13:
```suggestion
* Added ``--request, -r`` to allow specifying the raw request body without extra processing as an alternative to ``stdin``. (`#534`_)
```

BoboTiG:
@jakubroztocil we are good to go :)

Finally I did not use the short version `-r` as `--raw` is small to type; and it could be used for another option in the future.

jakubroztocil (changes requested):
Small docs tweaks

BoboTiG:
I'll backport changes to HTTP Prompt when this will get merged.

Also, if you think `-r` is still revelant here, let me know ;)

BoboTiG:
Good catch! You are right, `POST` should be used here to keep consistency with empty `stdin`. I'll fix it right now.

BoboTiG:
That's fixed :)

jakubroztocil:
Thanks, @BoboTiG 🎉

PR #1069, opened 2021-05-26
Modernize the code base with f-strings in tests

BoboTiG:
The code base is now using `f-string` almost everywhere instead of a mix of `str.format()`, `str + str + str` and `'%s...%d' % (arg1, arg2)`.

Simple concatenations were kept for readability purpose.

### DIFF
--- a/tests/utils/matching/test_matching.py
+++ b/tests/utils/matching/test_matching.py
@@ -15,7 +15,7 @@ def test_assert_output_matches_headers_unterminated():
     assert_output_does_not_match(
         (
             f'HTTP/1.1{CRLF}'
-            f'AAA:BBB'
+            'AAA:BBB'
             f'{CRLF}'
         ),
         [Expect.RESPONSE_HEADERS],
@@ -90,7 +90,7 @@ def test_assert_output_matches_headers_and_body():
             f'HTTP/1.1{CRLF}'
             f'AAA:BBB{CRLF}'
             f'{CRLF}'
-            f'CCC'
+            'CCC'
         ),
         [Expect.RESPONSE_HEADERS, Expect.BODY]
     )
@@ -121,7 +121,7 @@ def test_assert_output_matches_multiple_messages():
             f'EEE:FFF{CRLF}'
             f'{CRLF}'
 
-            f'GGG'
+            'GGG'
             f'{MESSAGE_SEPARATOR}'
         ), [
             Expect.REQUEST_HEADERS,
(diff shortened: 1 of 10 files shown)


jakubroztocil:
I’m a big fan of f-strings, but when it comes to a simple concatenation of variables or just appending something to a prefix (as opposed to variables appearing in the middle of a string), then `+` is often the more practical choice because that makes the code easier to write/read, and the intent clearer. For example, compare the following two snippets: 

```python
httpbin + '/post'
USERNAME + SEPARATOR_CREDENTIALS + PASSWORD
````

```python
f'{httpbin.url}/post'
f'{USERNAME}{SEPARATOR_CREDENTIALS}{PASSWORD}'
```

Also, even though it was used at some places, we don’t need the `.url` after `httpbin` / `httpbin_both`. I’d switch to the version without it everywhere.

BoboTiG:
Actually, those 2 lines produce different results:
```python
f'{httpbin}/post'
f'{httpbin.url}/post'
```
`httpbin + '/post'` works thanks to https://github.com/kevin1024/pytest-httpbin/blob/be3c3205972195ca632949ed7af1dc7086cb7899/pytest_httpbin/serve.py#L114-L115.

But when using `f-string`, we need to use the `.url` attribute.

BoboTiG:
OK, I will revert to simple `+` concatenations :+1:

jakubroztocil:
Good point about `httpbin.__add__()`.

BoboTiG:
The PR is now smaller.
Do not hesitate to nick-pick ;)

tests/utils/matching/test_matching.py
```diff
@@ -185,6 +185,6 @@ def test_assert_output_matches_multipart_body_with_separator():
 
 def test_assert_output_matches_multiple_separators():
     assert_output_matches(
-        MESSAGE_SEPARATOR + MESSAGE_SEPARATOR + 'AAA' + MESSAGE_SEPARATOR + MESSAGE_SEPARATOR,
+        f'{MESSAGE_SEPARATOR}{MESSAGE_SEPARATOR}AAA{MESSAGE_SEPARATOR}{MESSAGE_SEPARATOR}',
```
jakubroztocil, line 188: Here I prefer the original version. I find it much easier to scan :)

jakubroztocil (changes requested):

tests/utils/matching/test_matching.py
BoboTiG: That's right :+1:

jakubroztocil:
🙌🏻

PR #1070, opened 2021-05-27
Polish Python 2 removal

BoboTiG:


httpie/downloads.py
```diff
@@ -131,7 +128,7 @@ def filename_from_url(url: str, content_type: Optional[str]) -> str:
         else:
             ext = mimetypes.guess_extension(content_type)
 
-        if ext == '.htm':  # Python 3
```
BoboTiG, line 134: I am not sure about that comment: does it mean you wanted to keep track for later removal of the whole code block?

jakubroztocil:
Nice clean-up, @BoboTiG 🥧

httpie/downloads.py
jakubroztocil: I was to note that only Python 3, but Python not 2, seems to return `.htm`.

PR #1071, opened 2021-05-27
pytest: Add hidden files to norecursedirs

hroncok:
The default value already contains this,
but when setting a custom one, it was overridden.

In Fedora, we build the package in `.pyproject-builddir` and not ignoring it confuses pytest:

    _pytest.pathlib.ImportPathMismatchError: ('httpie.__main__', '/builddir/build/BUILD/httpie-2.4.0/.pyproject-builddir/pip-req-build-aedma65c/build/lib/httpie/__main__.py', PosixPath('/builddir/build/BUILD/httpie-2.4.0/.pyproject-builddir/pip-req-build-aedma65c/httpie/__main__.py'))

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,7 @@
 
 [tool:pytest]
 # <https://docs.pytest.org/en/latest/customize.html>
-norecursedirs = tests/fixtures
+norecursedirs = tests/fixtures .*
 addopts = --tb=native --doctest-modules
 
 


jakubroztocil:
Thanks for PR, @hroncok! 🎉

Is there any way we can help with the packaging for Fedora to make sure HTTPie is always up-to-date there? 

// cc @BoboTiG

hroncok:
I've just adopted the package, trying to enable tests on build time. Except one more PR shortly.

jakubroztocil:
That’s great to hear! @BoboTiG and I are here to help and answer any questions.

hroncok:
I've updated httpie in Fedora Rawhide (development version) to the latest version. I intend to update it in the recent stable Fedora release Fedora 34 as well. Fedora 33 is on 1.0.3 and I am not sure I can update to 2.x, will need to check how much backwards compatible it is.

jakubroztocil:
@hroncok Sweet! The biggest backward-incompatible change since `1.0.3` is that  `2.0.0` dropped Python 2.7 support. 

https://github.com/httpie/httpie/blob/master/CHANGELOG.rst

hroncok:
We run on Python 3 anyway, so that does not affect Fedora.

jakubroztocil:
Does the Fedora Rawhide package get propagated to RHEL/CentOS?

hroncok:
Definitively not.  RHEL/CentOS doe snot contain httpie.

jakubroztocil:
Is there anything we can do for httpie to be included in RHEL/CentOS?

hroncok:
A customer request might work for RHEL :money_mouth_face: 

An old version of httpie is actually included in [EPEL](https://fedoraproject.org/wiki/EPEL) 7. It is missing from EPEL 8.

I got a co-maintainer who's interested in maintaining the package in EPEL but frankly I am personally very hesitant in maintaining packages there because there are much harder requirements in API stability (e.g. wrt command line options etc.) and the task requires taking responsibility of something for a decade. The nature of (RH|EP)EL does not favor rolling updates and backporting (e.g. [security](https://bugzilla.redhat.com/show_bug.cgi?id=1751042)) bugfixes for 10+ years is not something I can pledge myself for :cry:

PR #1072, opened 2021-05-27
Skip http://pie.dev tests when offline

hroncok:


BoboTiG (approved):
Clever, thanks!

tests/conftest.py
```diff
@@ -22,3 +23,22 @@ def httpbin_secure_untrusted(monkeypatch, httpbin_secure):
...
+        return True
+    except OSError:
+        return False
+
+
+@pytest.fixture(scope='function')
+def httpbin_with_chunked_support(_httpbin_with_chunked_support_available):
+    if _httpbin_with_chunked_support_available:
+        return 'http://pie.dev'
+    pytest.skip('pie.dev not resolvable')
```
jakubroztocil, line 44: Could we change this so that `pie.dev` is not hard-coded but rather taken from the `HTTPBIN_WITH_CHUNKED_SUPPORT` constant?

jakubroztocil (changes requested):
Nice solution! Just one tweak needed.

tests/conftest.py
hroncok: Sure, on it.

hroncok: Done.

PR #1073, opened 2021-05-27
Remove an useless shebang form non-executable file

hroncok:
Shebangs have no function in non-executable files.
This file does not need to be directly executed.


hroncok:
This is probably the last one.

BoboTiG (approved):
Ah, it was on my todo list, good catch :+1:

jakubroztocil:
🙌🏻

PR #1076, opened 2021-05-27
Updated fish shell completions

exploide:
Just brought the completions for fish shell up to date.

### DIFF
--- a/extras/httpie-completion.fish
+++ b/extras/httpie-completion.fish
@@ -1,60 +1,137 @@
-function __fish_httpie_auth_types
-  echo "basic"\t"Basic HTTP auth"
-  echo "digest"\t"Digest HTTP auth"
-end
-
 function __fish_httpie_styles
-  echo "autumn"
-  echo "borland"
-  echo "bw"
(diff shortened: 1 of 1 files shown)


jakubroztocil:
Thanks, @exploide! 🐟🐚 🎉

hroncok:
I was jst installing those in the Fedora package and was wondering: Will this work for `https` command?

exploide:
> I was just installing those in the Fedora package

Hey @hroncok thanks for addressing my bugzilla issue. :) Just left positive karma on bodhi for the F34 update.

As of now, I was not even aware that there is an `https` command (that is the same as `http`).

> Will this work for https command?

Not directly. We would need an additional completion file `https.fish` that just contains `complete -c https -w http`. This means that completions for `https` are wrapping those of `http`.

How should be proceed? Will you create that file as part of the build step? Or we can ask @jakubroztocil if we can add that additional file to the repository.

hroncok:
Having that wrapper here would make more sense to me, that way, others can use it as well.
Is something similar working for Bash?

PR #1078, opened 2021-05-29
Minor clean-up

BoboTiG:
- Remove default arguments to `open()`.
- Make use of `pytest` mechanisms for temporary folders.


jakubroztocil:
Cheers, @BoboTiG 🎉

PR #1080, opened 2021-05-31
Review OSError exceptions handling

BoboTiG:
- Replace obsolete `IOError` (Python 2) with `OSError`, cf https://docs.python.org/3/library/exceptions.html#OSError.
- Improve `OSError` catches at different places, simplifying the code.


jakubroztocil:
Thanks! 🙌🏻

PR #1081, opened 2021-06-01
Fix several issues found with flake8

BoboTiG:


setup.cfg
```diff
@@ -17,7 +17,13 @@ addopts = --tb=native --doctest-modules
...
 # W503 - line break before binary operator
-ignore = E241,E501,W503
+ignore = E501,W503
+
+
+[flake8]
+# <https://flake8.pycqa.org/en/latest/user/error-codes.html>
+# E501 - line too long
+# W503 - line break before binary operator
+ignore = E501,W503
```
jakubroztocil, line 29: Why both pycodestyle and flake8?

BoboTiG: `pycodestyle` does not catch `flake8` errors (error codes starting with `F`).
That section in `setup.cfg` is useful if we want to use `flake8` as well as `pycodestyle` (it would have been a task for upcoming days).

I can remove that section and just commit fixes.

---

FTR here are missed issues by `pycodestyle`:
```
F401 'xxx' imported but unused
F541 f-string is missing placeholders
F841 local variable 'r' is assigned to but never used
```
And using some `flake8` extensions, we could even find such issue:
```
C413 Unnecessary list call around sorted()
```

We have several options:
1. use only `pycodestyle`
2. use only `flake8`
3. use both `pycodestyle` and `flake8`

I do not have a strong opinion, since that PR fixes all remaining problems found by `flake8`.

BoboTiG: Actually, it might just be a `pycodestyle` configuration to tweak to support all error codes. I'll come back to it later.

jakubroztocil: It would be good to settle on one tool, if possible. Currently, `pycodestyle` configured to be invoked via `make` and in the GitHub action.

jakubroztocil:
🧹🎉

setup.cfg
BoboTiG: Yup, I'll do it in the coming days.

PR #1084, opened 2021-06-06
Warn against non-ascii in setup.cfg

OsvaldoRino:


BoboTiG (approved):

jakubroztocil:
What is the actual change and the reason behind it here? The current version looks identical:

https://github.com/httpie/httpie/blob/350f973f705dc996af27d66fe02e03e2f1f99968/setup.cfg#L1-L5

PR #1086, opened 2021-06-09
Add a Packit configuration for Fedora packaging

hroncok:
- all pull requests are build-tested in https://copr.fedorainfracloud.org
 - new releases will create pull requests in https://src.fedoraproject.org/rpms/httpie

As discussed with @BoboTiG over email. Not tested yet.

The CI build will fail because of https://github.com/httpie/httpie/pull/1057#issuecomment-857684958 -- once that is fixed, I'll rebase this.

The service needs to be enabled first: https://packit.dev/docs/packit-as-a-service/

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,19 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+specfile_path: httpie.spec
+actions:
+  # the current Fedora Rawhide specfile has some patches
+  # so we get it from @hroncok's (= churchyard in Fedora) fork for now
+  # once we have a new release, we'll use: https://src.fedoraproject.org/rpms/httpie/raw/rawhide/f/httpie.spec
+  post-upstream-clone: "wget https://src.fedoraproject.org/fork/churchyard/rpms/httpie/raw/packit/f/httpie.spec -O httpie.spec"
+jobs:
+- job: copr_build
+  trigger: pull_request
+  metadata:
+    targets:
+    - fedora-all
+- job: propose_downstream
+  trigger: release
+  metadata:
+    dist_git_branches:
+    - rawhide
--- a/.gitignore
+++ b/.gitignore
@@ -139,3 +139,8 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Packit
+/httpie.spec
+/httpie-*.rpm
+/httpie-*.tar.gz
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,7 @@ setup(
     author=httpie.__author__,
     author_email='jakub@roztocil.co',
     license=httpie.__licence__,
-    packages=find_packages(),
+    packages=find_packages(include=['httpie', 'httpie.*']),
     entry_points={
         'console_scripts': [
             'http = httpie.__main__:main',


.packit.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+specfile_path: httpie.spec
+synced_files:
+  - .packit.yaml
+upstream_package_name: httpie
+downstream_package_name: httpie
+actions:
+  # the current Fedora Rawhide specfile has some patches
+  # so we get it from @hroncok's (= churchyard in Fedora) fork for now
+  # once we have a new relese, we'll use: https://src.fedoraproject.org/rpms/httpie/raw/rawhide/f/httpie.spec
```
BoboTiG, line 12: Nitpick:
```suggestion
  # once we have a new release, we'll use: https://src.fedoraproject.org/rpms/httpie/raw/rawhide/f/httpie.spec
```

BoboTiG (commented):
That is awesome, thanks!

If I understand well, there is no more need for us to provide the `httpie.spec` file because the one from https://src.fedoraproject.org/rpms/httpie/raw/rawhide/f/httpie.spec will be used, right?

hroncok:
> If I understand well, there is no more need for us to provide the httpie.spec file because the one from https://src.fedoraproject.org/rpms/httpie/raw/rawhide/f/httpie.spec will be used, right?

Indeed. As a downstream maintainer, I'd like to avoid keeping the specfile in upstream, as it makes my life harder than necessary.

BoboTiG:
And it makes mine easier, so it is perfect :)

BoboTiG:
Packit GH app is installed ;)

hroncok:
> Namespace is not allowed! 

Let me see... :/

hroncok:
https://github.com/packit/packit-service/issues/997 seems to indicate that "Namespace is not allowed!" means that some real human from the packit team will get notified that we attempted to use it here and they need to allow github.com/httpie manually to verify we are not scumbags. Is that correct @TomasTomecek?

hroncok:
> [packit/packit-service#997](https://github.com/packit/packit-service/issues/997) seems to indicate that "Namespace is not allowed!" means that some real human from the packit team will get notified that we attempted to use it here and they need to allow github.com/httpie manually to verify we are not scumbags.

Fixed in https://github.com/packit/notifications/issues/361

hroncok:
It appears to work. The build failures are real and they are related to https://github.com/httpie/httpie/pull/1057#issuecomment-857684958

BoboTiG:
Thanks a lot @hroncok !

jakubroztocil:
Awesome stuff 📦🎉🚀

TomasTomecek:
> [packit/packit-service#997](https://github.com/packit/packit-service/issues/997) seems to indicate that "Namespace is not allowed!" means that some real human from the packit team will get notified that we attempted to use it here and they need to allow github.com/httpie manually to verify we are not scumbags. Is that correct @TomasTomecek?

you are correct, we had to put this namespace on a allowlist :)

https://packit.dev/docs/faq/#can-i-use-the-packit-service-as-soon-as-i-install-it-into-my-repository

sorry for a late reply - summer is coming :) finally tracking down all the notifications

@hroncok thank you for setting this up!!

PR #1087, opened 2021-06-09
Prevent packaging of test files

BoboTiG:
cc @hroncok

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,7 @@ setup(
     author=httpie.__author__,
     author_email='jakub@roztocil.co',
     license=httpie.__licence__,
-    packages=find_packages(),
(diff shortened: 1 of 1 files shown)


hroncok:
This should fix the problem. To verify, I've included this commit in https://github.com/httpie/httpie/pull/1086

hroncok:
Oh, no. It does not include subpackages :(

BoboTiG:
Yup I just fixed it now :)

hroncok:
I'd use `setuptools.find_packages(include=['httpie*'])` to avoid a hardcoded list.

jakubroztocil:
Thanks @BoboTiG & @hroncok! 🎉🥧

PR #1089, opened 2021-06-11
Use a more modern approach to run tests

BoboTiG:
Running tests through `python setup.py test` is deprecated:

> WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.

I am not in favor of moving back to `tox`, we should simply run tests using `python -m pytest` (or `make test`) and that's it.

A new extra was added, `dev`, to install development requirements:
```shell
$ python -m pip install --upgrade --editable '.[dev]'
```


.github/workflows/build.yml
```diff
@@ -30,5 +31,5 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - run: python -m pip install --upgrade pip setuptools wheel
-      - run: python -m pip install --upgrade --editable .
-      - run: python setup.py test
+      - run: python -m pip install --upgrade --editable . -r requirements-dev.txt
+      - run: python -m pytest --verbose ./httpie ./tests
```
BoboTiG, line 35: FTR `--doctest-modules` is missing here because it is already defined in `setup.cfg`.

BoboTiG:
Sorry to ping you @hroncok :)

According to https://src.fedoraproject.org/rpms/httpie/blob/rawhide/f/httpie.spec#_40, does the failure seems legit to you?

> RPM build errors:
> /var/tmp/rpm-tmp.4j0VHh: line 33: /usr/bin/pytest: No such file or directory

Do you see somethng bad with those changes?

hroncok:
https://src.fedoraproject.org/rpms/httpie/blob/rawhide/f/httpie.spec#_40 was removed in the forked spec we use in favor of https://src.fedoraproject.org/fork/churchyard/rpms/httpie/blob/packit/f/httpie.spec#_31

will comment inline

.github/workflows/build.yml
```diff
@@ -30,5 +31,5 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - run: python -m pip install --upgrade pip setuptools wheel
-      - run: python -m pip install --upgrade --editable .
-      - run: python setup.py test
+      - run: python -m pip install --upgrade --editable . -r requirements-dev.txt
```
hroncok, line 34: The `[test]` extra can also be used here and in any instructions for running tests.

```suggestion
      - run: python -m pip install --upgrade --editable .[test] -r requirements-dev.txt
```

setup.py
```diff
@@ -59,7 +30,6 @@ def run_tests(self):
 
 # bdist_wheel
 extras_require = {
-    'test': tests_require,
```
hroncok, line 62: The Fedora package uses the `[test]` extra to fetch test dependencies. I'd appreciate if this could stay.

BoboTiG:
Wait before doing the change :)
Do you recommand using another way to specify requirements?

.github/workflows/build.yml
hroncok: All the other requirements from `requirements-dev.txt` could be defined as `[dev]` extra and the command could be:

```
python -m pip install --upgrade --editable .[test,dev]
```

I'd also advise against using `--editable` on the CI. It is mostly also semi-deprecated / things are broken with PEP517/PEP518.

setup.py
BoboTiG: Sure, I'll revert that change.

.github/workflows/build.yml
BoboTiG: Do you mean to simply use
```
python -m pip install --upgrade .[test,dev]
```
?

hroncok: yes

setup.py
```diff
@@ -4,39 +4,26 @@
...
-
+# Note: keep requirements here to ease distributions packaging
 tests_require = [
-    'pytest-httpbin',
     'pytest',
+    'pytest-httpbin>=0.0.6',
+]
+dev_require = [
+    *tests_require,
+    'docutils',
```
hroncok, line 17: IIRC docutils could be moved to tests_require, there is a test that uses them.

.github/workflows/build.yml
BoboTiG: Thanks, I learned something today :)

jakubroztocil:
🥧

PR #1094, opened 2021-06-19
Support requests.response.raw being a file-like object

IlyaSukhanov:
Previously httpie relied on requests.models.Response.raw being
urllib3.HTTPResponse.  The Requests documentation specifies that
[requests.models.Response.raw](https://docs.python-requests.org/en/master/api/#requests.Response.raw)
is a File-like object but allows for other types for internal use.

This change introduces graceful handling for scenarios when
requests.models.Response.raw is not urllib3.HTTPResponse. In such a scenario
httpie now falls back to extracting metadata from requests.models.Response
directly instead of direct access from protected protected members such as
response.raw._original_response. A side effect in this fallback procedure is
that we can no longer determine HTTP protocol version and report it as `??`.

This change is necessary to make it possible to implement TransportPlugins
without having to also needing to emulate internal behavior of urlib3 and
http.client.

This pull requests is an attempt to address https://github.com/httpie/httpie/issues/1093

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
   an alternative to ``stdin``. (`#534`_)
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
 * Fixed ``--verbose`` HTTP 307 redirects with streamed request body. (`#1088`_)
+* Fixed support of file-like object responses. (`#1094`_)
 
 
 `2.4.0`_ (2021-02-06)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
   an alternative to ``stdin``. (`#534`_)
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
 * Fixed ``--verbose`` HTTP 307 redirects with streamed request body. (`#1088`_)
-* Fixed support of file-like object responses. (`#1094`_)
+* Add internal support for file-like object responses to improve adapter plugin support. (`#1094`_)
 
 
 `2.4.0`_ (2021-02-06)
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -104,10 +104,15 @@ def collect_messages(
                     **send_kwargs,
                 )
 
-            # noinspection PyProtectedMember
-            expired_cookies += get_expired_cookies(
-                headers=response.raw._original_response.msg._headers
-            )
+            if isinstance(response.raw, urllib3.HTTPResponse):
+                # noinspection PyProtectedMember
+                expired_cookies += get_expired_cookies(
+                    headers=response.raw._original_response.msg._headers
+                )
+            else:
+                expired_cookies += get_expired_cookies(
+                    headers=response.headers.items()
+                )
 
             response_count += 1
             if response.next:
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -104,15 +104,9 @@ def collect_messages(
                     **send_kwargs,
                 )
 
-            if isinstance(response.raw, urllib3.HTTPResponse):
-                # noinspection PyProtectedMember
-                expired_cookies += get_expired_cookies(
-                    headers=response.raw._original_response.msg._headers
-                )
-            else:
-                expired_cookies += get_expired_cookies(
-                    headers=response.headers.items()
-                )
+            expired_cookies += get_expired_cookies(
+                response.headers.get('Set-Cookie', '')
+            )
 
             response_count += 1
             if response.next:
--- a/httpie/models.py
+++ b/httpie/models.py
@@ -1,5 +1,6 @@
 from typing import Iterable, Optional
 from urllib.parse import urlsplit
+import urllib3
 
 
 class HTTPMessage:
@@ -49,9 +50,8 @@ class HTTPResponse(HTTPMessage):
     def iter_lines(self, chunk_size):
         return ((line, b'\n') for line in self._orig.iter_lines(chunk_size))
 
-    # noinspection PyProtectedMember
-    @property
-    def headers(self):
+    def _headers_http_client_http_message(self):
+        """ Extract header information from http.client.HTTPMessage """
         original = self._orig.raw._original_response
 
         version = {
@@ -61,12 +61,25 @@ class HTTPResponse(HTTPMessage):
             20: '2',
         }[original.version]
 
-        status_line = f'HTTP/{version} {original.status} {original.reason}'
+        return version, original.status, original.reason, original.msg._headers
+
+    def _headers_requests_response(self):
+        """ Extract header information from Requests.models.Response """
+        version = '1.1'
+        return version, self._orig.status_code, self._orig.reason, self._orig.headers.items()
+
+    @property
+    def headers(self):
+        if isinstance(self._orig.raw, urllib3.HTTPResponse):
+            header_extract_method = self._headers_http_client_http_message
+        else:
+            header_extract_method = self._headers_requests_response
+        version, status, reason, http_headers = header_extract_method()
+
+        status_line = f'HTTP/{version} {status} {reason}'
         headers = [status_line]
-        # `original.msg` is a `http.client.HTTPMessage`
-        # `_headers` is a 2-tuple
         headers.extend(
-            f'{header[0]}: {header[1]}' for header in original.msg._headers)
+            ': '.join(header) for header in http_headers)
         return '\r\n'.join(headers)
 
     @property
--- a/httpie/models.py
+++ b/httpie/models.py
@@ -1,6 +1,8 @@
 from typing import Iterable, Optional
 from urllib.parse import urlsplit
 
+from httpie.utils import split_cookies
+
 
 class HTTPMessage:
     """Abstract class for HTTP messages."""
@@ -68,7 +70,14 @@ class HTTPResponse(HTTPMessage):
         status_line = f'HTTP/{version} {original.status_code} {original.reason}'
         headers = [status_line]
         headers.extend(
-            ': '.join(header) for header in original.headers.items())
+            ': '.join(header)
+            for header in original.headers.items()
+            if header[0] != "Set-Cookie"
+        )
+        headers.extend(
+            f'Set-Cookie: {cookie}'
+            for cookie in split_cookies(original.headers.get('Set-Cookie'))
+        )
         return '\r\n'.join(headers)
 
     @property
--- a/httpie/models.py
+++ b/httpie/models.py
@@ -49,29 +49,26 @@ class HTTPResponse(HTTPMessage):
     def iter_lines(self, chunk_size):
         return ((line, b'\n') for line in self._orig.iter_lines(chunk_size))
 
-    def _headers_extract_version(self):
+    # noinspection PyProtectedMember
+    @property
+    def headers(self):
         try:
             raw_version = self._orig.raw._original_response.version
         except AttributeError:
-            raw_version = None
-        return {
+            # Assume HTTP/1.1
+            raw_version = '1.1'
+        version = {
             9: '0.9',
             10: '1.0',
             11: '1.1',
             20: '2',
-        }.get(raw_version, '1.1')
+        }[raw_version]
 
-    @property
-    def headers(self):
-        version = self._headers_extract_version()
-        status = self._orig.status_code
-        reason = self._orig.reason
-        http_headers = self._orig.headers.items()
-
-        status_line = f'HTTP/{version} {status} {reason}'
+        original = self._orig
+        status_line = f'HTTP/{version} {original.status_code} {original.reason}'
         headers = [status_line]
         headers.extend(
-            ': '.join(header) for header in http_headers)
+            ': '.join(header) for header in original.headers.items())
         return '\r\n'.join(headers)
 
     @property
--- a/httpie/models.py
+++ b/httpie/models.py
@@ -1,6 +1,5 @@
 from typing import Iterable, Optional
 from urllib.parse import urlsplit
-import urllib3
 
 
 class HTTPMessage:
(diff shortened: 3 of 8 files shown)


httpie/models.py
```diff
@@ -61,12 +61,27 @@ def headers(self):
             20: '2',
         }[original.version]
 
-        status_line = f'HTTP/{version} {original.status} {original.reason}'
+        return version, original.status, original.reason, original.msg._headers
```
IlyaSukhanov, line 64: Status, reason and headers ought to be pulled from self._orig as is done in _headers_requests_response instead of accessing the protected member self._orig.raw._original_response. Its not clear why this was not done to begin with and for that reason I'm leaving the old logic in place.

Notable that same approach cannot be used for version. Perhaps it would be better to file a bug against requests and ask for HTTP version to be exposed without having to rely on protected members.

httpie/client.py
```diff
@@ -104,10 +104,15 @@ def collect_messages(
...
                 )
 
-            # noinspection PyProtectedMember
-            expired_cookies += get_expired_cookies(
-                headers=response.raw._original_response.msg._headers
-            )
+            if isinstance(response.raw, urllib3.HTTPResponse):
+                # noinspection PyProtectedMember
+                expired_cookies += get_expired_cookies(
+                    headers=response.raw._original_response.msg._headers
```
IlyaSukhanov, line 110: Its not entirely clear to me why `response.raw._original_response.msg._headers` is being utilized here instead of `response.headers`. I did run tests without making use of  `response.raw._original_response.msg._headers` and they passed. Logic is left in place to ensure compatibility for what ever case this was intended for, however would be happy to remove on recommendation from someone more experienced with the project.

IlyaSukhanov: Upon further reflection and re-reading https://github.com/httpie/httpie/pull/929#discussion_r439017110 its probably because Requests represents headers in a dict instead of a tuple list. Headers with same name would be lost.

IlyaSukhanov: Upon further experimentation it turns out its possible to not make use of protected _original for purposes of get_expired_cookies(). Here is an passing implementation https://github.com/IlyaSukhanov/httpie/commit/36daab668c8357d61dd27a954ab8aae556eca202. Can add this commit to this PR if desired.

BoboTiG: The last iteration on your branch seems reasonable :+1: 
Could you merge it there please?

IlyaSukhanov: Done

httpie/models.py
IlyaSukhanov: If its reasonable to assume that
```
status = self._orig.raw._original_response.status_code
reason = self._orig.raw._original_response.reason
headers = self._orig.raw._original_response.msg._headers
```
Is effectively the same as.
```
status = self._orig.status_code
reason = self._orig.reason
headers = self._orig.headers.items()
```

Then its worthwhile to consider the slightly simpler implementation https://github.com/IlyaSukhanov/httpie/commit/c14deea1a68c979f7700be1217512ee39c06d742

BoboTiG: I would be in favor of the slightly simpler implementation but I would like @jakubroztocil POV, if possible.

httpie/utils.py
```diff
@@ -96,16 +99,17 @@ def is_expired(expires: Optional[float]) -> bool:
...
     )
+
     cookies = [
         # The first attr name is the cookie name.
         dict(attrs[1:], name=attrs[0][0])
         for attrs in attr_sets
     ]
 
     _max_age_to_expires(cookies=cookies, now=now)
+    "Set-Cookie: foo=bar; Expires=Mon, 21 Jun 2021 04:20:21 GMT; Max-Age=1; Path=/"
```
BoboTiG, line 112: Is it a comment left for debugging or to expose the cookie syntax?

BoboTiG (commented):
That seems promising :muscle: 

Let's wait on Jakub for one specific area of the patch.
In the meantime, could you stick with the project quotes guidelines (`"` -> `'`)?

httpie/utils.py
IlyaSukhanov: Note to self that I forgot to delete.

httpie/models.py
```diff
@@ -61,12 +61,25 @@ def headers(self):
...
+
+    def _headers_requests_response(self):
+        """ Extract header information from Requests.models.Response """
+        version = "??"
```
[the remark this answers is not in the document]
IlyaSukhanov: Seems like a reasonable default. Done!

httpie/models.py
```diff
@@ -61,12 +61,25 @@ def headers(self):
...
+            header_extract_method = self._headers_requests_response
+        version, status, reason, http_headers = header_extract_method()
+
+        status_line = f'HTTP/{version} {status} {reason}'
         headers = [status_line]
-        # `original.msg` is a `http.client.HTTPMessage`
-        # `_headers` is a 2-tuple
         headers.extend(
-            f'{header[0]}: {header[1]}' for header in original.msg._headers)
+            ": ".join(header) for header in http_headers)
```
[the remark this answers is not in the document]
IlyaSukhanov: quotes fixed

httpie/models.py
IlyaSukhanov: Sounds good. Let me know if you'd prefer I open that as a separate pr or include it on this one.

IlyaSukhanov:
Thank for for review and feedback.

Latest changes:
* fixed quotes for consistency
* in unknown case http protocol version is set to 1.1
* removed stray debug comment

httpie/models.py
BoboTiG: Let's go with the simplified version :+1:

Makefile
```diff
@@ -77,14 +77,20 @@ venv:
 
 test:
 	@echo $(H1)Running tests$(HEADER_EXTRA)$(H1END)
-	$(VENV_BIN)/python -m pytest $(COV) ./httpie $(COV) ./tests --doctest-modules --verbose ./httpie ./tests
+	$(VENV_BIN)/python -m pytest $(COV) ./httpie $(COV) ./tests --doctest-modules --verbose ./httpie ./tests $(MISSING)
```
BoboTiG, line 80: Could you revert those changes and propose them in a specific PR if you think it is interesting to keep :pray: ?

IlyaSukhanov: certainly

httpie/models.py
IlyaSukhanov: Done

Makefile
IlyaSukhanov: Done

IlyaSukhanov:
* Make file alterations have been reverted.
* Simplified envelope extraction added.

httpie/models.py
```diff
@@ -52,21 +52,23 @@ def iter_lines(self, chunk_size):
     # noinspection PyProtectedMember
     @property
     def headers(self):
-        original = self._orig.raw._original_response
-
+        try:
+            raw_version = self._orig.raw._original_response.version
+        except AttributeError:
+            # Assume HTTP/1.1
+            raw_version = '1.1'
```
BoboTiG, line 59:
```suggestion
            raw_version = 11
```

BoboTiG (approved):

tests/test_sessions.py
```diff
@@ -368,11 +363,10 @@ def test_get_expired_cookies_using_max_age(self):
...
             (
-                [
-                    ('Set-Cookie', 'hello=world; Path=/; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly'),
-                    ('Set-Cookie', 'pea=pod; Path=/ab; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly'),
-                    ('Connection', 'keep-alive')
-                ],
+                (
+                    'hello=world; Path=/; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly, '
+                    'pea=pod; Path=/ab; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly'
+                ),
```
jakubroztocil, line 369: These should now be a `list`s.

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
   an alternative to ``stdin``. (`#534`_)
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
 * Fixed ``--verbose`` HTTP 307 redirects with streamed request body. (`#1088`_)
+* Fixed support of file-like object responses. (`#1094`_)
```
jakubroztocil, line 16:
```suggestion
* Add internal support for file-like object responses to improve adapter plugin support. (`#1094`_)
```

jakubroztocil (changes requested):

tests/test_sessions.py
IlyaSukhanov: Cookies were stored as a list in `response.raw._original_response.msg._headers` however with this change cookies are extracted from `response.headers.get('Set-Cookie', '')` which is a comma separated string.

get_expired_cookies has been changed from accepting list-of-tuple headers to string of comma separated strings.

I don't see of a way to access more-or-less raw cookie data from requests in any way but parsing them out of the comma separated string without resorting to using protected members. Cookie jar has more formatted cookies but they are stored in a lossy way.

With all this in mind, to keep tests as lists the cookie splitting logic can be moved to outside of `get_expired_cookies()`, but I'm not sure if this is what you had in mind by your comment.

Thanks
--IAS

BoboTiG: @IlyaSukhanov Is it an option to add a test case to ensure `response.headers.get('Set-Cookie', '')` returns a string with concatenated cookies? It will help us prevent potential regressions for upcoming `requests` versions and to be more comfortable with those changes too :)

After that, we will be good to merge.

IlyaSukhanov: Added test to ensure that multiple cookies get extracted correctly.

IlyaSukhanov: Looks like cicd checks is failing. Is there a tried approach for starting a local http server to test against?

The way it's erroring out it's not clear what the issue is. Few of suspicions I have:

1. Multiprocess does not start for some reason
2. Multiprocess does not start fast enough (single threaded machine?)
3. Flask is unable to open port for some reason (seems unlikely as no error to that effect is presented)


Without knowing what the error is or much experience with GitHub check infrastructure. My only idea is to add a retry functionality in test, to call httpie multiple time until hopefully the background server becomes responsive.

BoboTiG (commented):
Is there a way to craft the `Response` without using multiprocessing and `flask`?

IlyaSukhanov:
If the objective is to test internal of requests to ensure that it returns cookies in right format (and does not regress in future), the viable options I can think of are:

1. Call a local http server and have it respond with desired payload. This is what e57adec does. e57adec makes use of flask and multiprocessing to start the local http server. There are other options besides flask (`http.server.HTTPServer`) for this task, but it would still need to rely on multiprocess to ensure server and request can run in parallel.. and it would possibly be sort of fragile unless of course there are ready made solutions made for this that are thoroughly tested.
2. Mock downstream of `requests`; either `urllib3` or `socket` and pass up a fake payload through requests ensuring request handles it correctly. I suspect this is not a trivial implementation but could be made easier via something like https://github.com/mindflayer/python-mocket

Its also possible to mock `requests.Response` directly but it would not protect from internal implementation changes in requests down the road.

BoboTiG:
Maybe something "lighter" like done in https://github.com/getsentry/responses/blob/bdf4c3b1273025338c6a475a556d741552ffa819/responses/test_responses.py#L845 could be interesting?

I would love to see something using "raw" objects (from `urllib3`/`requests`/...) to simply tell the response there is several cookies. And then, check they all are processed as expected.

I do not have a strong opinion as soon as tests are green et possibly not too random.

tests/test_cookie.py
```diff
@@ -0,0 +1,37 @@
...
+    conn = socket()
+    conn.bind(('', 0))
+    port = conn.getsockname()[1]
+    conn.close()
+    return port
+
+
+def test_cookie_parser():
+    port = available_port()
+    server = Process(target=app.run, kwargs={'port': port})
```
BoboTiG, line 29:
```suggestion
    server = Process(target=app.run, kwargs={'port': 8888})
```

tests/test_cookie.py
```diff
@@ -0,0 +1,37 @@
...
+    conn = socket()
+    conn.bind(('', 0))
+    port = conn.getsockname()[1]
+    conn.close()
+    return port
+
+
+def test_cookie_parser():
+    port = available_port()
+    server = Process(target=app.run, kwargs={'port': 8888})
```
BoboTiG, line 29:
```suggestion
    port = 8080
    server = Process(target=app.run, kwargs={'port': port})
```

BoboTiG:
@IlyaSukhanov I'll handle the cookies test, it is a bonus and you've done so much already :+1:

setup.py
```diff
@@ -23,6 +23,7 @@
     'pytest-cov',
     'twine',
     'wheel',
+    'flask',
```
BoboTiG, line 26:
```suggestion
```

tests/test_cookie.py
```diff
@@ -0,0 +1,47 @@
...
+            """"HTTP request handler."""
+
+            def do_GET(self):
+                """Handle GET requests."""
+                # Craft multiple cookies
+                cookie = SimpleCookie()
+                cookie['hello'] = 'world'
+                cookie['hello']['path'] = '/'
+                cookie['oatmeal_raisin'] = 'is the best'
+                cookie['oatmeal_raisin']['path'] = '/'
```
BoboTiG, line 37:
```suggestion
                cookie['hello'] = 'world'
                cookie['hello']['path'] = self.path
                cookie['oatmeal_raisin'] = 'is the best'
                cookie['oatmeal_raisin']['path'] = self.path
```

BoboTiG:
OK it should be good now. Do you mind having a final look at the test @IlyaSukhanov?

httpie/models.py
```diff
@@ -1,6 +1,8 @@
 from typing import Iterable, Optional
 from urllib.parse import urlsplit
 
+from httpie.utils import split_cookies
```
BoboTiG, line 4:
```suggestion
from .utils import split_cookies
```

httpie/utils.py
```diff
@@ -85,8 +88,21 @@ def get_content_type(filename):
         return content_type
 
 
+def split_cookies(cookies):
+    """
+    When Requests stores cookies in ``response.headers['Set-Cookie']``
+    it concatenates all of them through ``, ``
```
BoboTiG, line 94:
```suggestion
    When ``requests`` stores cookies in ``response.headers['Set-Cookie']``
    it concatenates all of them through ``, ``.
```

IlyaSukhanov:
Elegant!

An observation, and not a problem: The thread is not actively terminated. I'm not sure if pytest kills threads of between each test. Even if it doesn't I cannot think of a scenario where current implementation would become problematic, if more tests of this sort were added each test server would be isolated to its own thread/port. And obviously pytest finishes execution so it must be terminating the thread at least at the completion of all tests.

:+1: :fireworks:

tests/test_sessions.py
jakubroztocil: I was referring to the incorrect usage of tuples for what by nature is a list. 

```python
(  # <= tuple because fixed-length collection where each item is a different thing

    ( # <= list because it’s a simple list of things
        'hello=world; Path=/; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly, '
        'pea=pod; Path=/ab; Expires=Thu, 01-Jan-1970 00:00:00 GMT; HttpOnly'
    ),
    None,
    [  # <= list because it’s a simple list of things
        {'name': 'hello', 'path': '/'},
        {'name': 'pea', 'path': '/ab'}
    ]
),
```

IlyaSukhanov: I think I see what you mean. But the cookie values are not tuples its a string that is line wrapped, and thus grouped using parenthesis.

Tuple:
```
(
    'foo',
    'bar'
)
```

String:
```
(
    'foo'
    'bar'
)
```

Note how in this example and in the pr the lines are not separated by comma which is required in tuple definition.

I hope this clears up any confusion.

If project has different style requirement for line wrapping long strings, please let me know and I'll make this change conform to that.

jakubroztocil: @IlyaSukhanov right, I see it now! Yeah, this style of string-wrapping is the preferred one, but in this context, next to a bunch of tuples, it’s easy to misread it (🙋‍♂️), so I’d vote for using a different style here.

BoboTiG:
Yes, about the zombie thread, it is not an issue right now. And I failed to find a multi-platform solution. It could be improved if we ever need more tests like that.

BoboTiG (approved):

BoboTiG:
That's awesome, thanks a lot for your work and patience @IlyaSukhanov :pie: :champagne:

PR #1102, opened 2021-07-02
Use echo with -n

marcelstoer:
I argue that you most likely don't want/need to send the trailing newline.


BoboTiG (approved):

BoboTiG:
Thanks @marcelstoer :pie:

PR #1103, opened 2021-07-02
Add the Snap build file for general Linux packaging

BoboTiG:
Based on the work of @elopio and @chipaca.

FTR the final package is 5.4 MB.

---

TODO:
- [x] Snap `http` and `https` [aliases](https://forum.snapcraft.io/t/request-aliases-for-httpie/25245) (edit: first need to merge that PR and upload our first package to the `httpie` repository).
- [x] get back the `snapd://` protocol support (edit: done with https://github.com/httpie/httpie-snapdsocket)
- [ ] an issue also present in the current unofficial package: the configuration file is not taken into account because of environment variables handling by `snapd`.

It would be good to also find a way to actually test those packages.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -57,7 +57,13 @@ $ port install httpie
 
 ### Linux
 
-Most Linux distributions provide a package that can be installed using the
+HTTPie is available on the [Snap Store](https://snapcraft.io/httpie):
+
+```bash
+$ snap install httpie
+```
+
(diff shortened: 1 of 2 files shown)


om26er:
@BoboTiG Looking at https://forum.snapcraft.io/t/request-aliases-for-httpie/25245/5, can you please upload a snap to the store with the current packaging. That will give us the required aliases

PR #1108, opened 2021-07-08
Extend plugins lookup paths

BoboTiG:
Plugins lookup paths are automatically extended to allow searching in several well-known locations. It will fix issues when HTTPie is installed through Homebrew and plugins through `pip`, on macOS for instance.

The fix is global and ease future additions to lookup paths.

Third-party applications can also extend the new module-level constant  `plugin.manager.EXTENDED_PLUGINS_SEARCH_AREA` to include more specific locations.

There is no new option as I think it will be overkill: installing modules in non-standard locations is something not recommanded and target a small amount of our users, if any.


BoboTiG:
I am wondering if the `EXTENDED_PLUGINS_SEARCH_AREA` should be filled on macOS only, and kept empty on other platforms :thinking:

jakubroztocil:
This looks like it solves the issue of plugins not being found when installed elsewhere, which is great. However, I see some potential systemic issues with this approach. 

They stem from the fact that plugins currently don’t have their own place. But instead, they reside in `site-packages` together with all the other installed Python packages. 

Correct me if my understanding is incorrect, but I’d like to know how you think about the following:

1. This doesn’t just search for plugin packages in those extra `site-packages` directories, but blank extends `sys.path`.
2. If some package HTTPie uses is missing from the interpreter's default `sys.path`, but one of those extra paths happens to contain it, it will be imported from there instead.
3. This can lead to unexpected and confusing situations.
4. This breaks the isolation provided by `venv` / `virtualenv`, complicating development.
5. It will be hard to reason where which plugin came from.
6. This might negatively impact startup time.

BoboTiG:
Yeah, all those questions are showing the bad decision in that PR :)
I will rethink it.

PR #1110, opened 2021-07-16
Improve handling of prettified responses without correct content-type encoding

### LINKED ISSUE #358: Default to encoding specified in XML declaration
HTTPie seems to assume ISO-8859-1 when decoding XML in UTF-8 form. Below I show the difference between the HTTPie and cURL outputs. The last element displays as “FinanciÃ«n | Begroting”, respectively “Financiën | Begroting” (Dutch for “Finance | Budget”).

$ http 'https://zoek.officielebekendmakingen.nl/kst-34200-14/metadata_owms.xml'

``` xml
HTTP/1.1 200 OK
Cache-Control: private
Connection: keep-alive
Content-Disposition: attachment; filename=metadata_owms.xml
Content-Encoding: gzip
Content-Type: text/xml
Date: Fri, 29 May 2015 09:14:39 GMT
Server: nginx
Set-Cookie: ASP.NET_SessionId=d5rme4n4tyrzt52z4hrto5fk; path=/; secure; HttpOnly
Transfer-Encoding: chunked
Vary: Accept-Encoding
X-Frame-Options: SAMEORIGIN

<?xml version="1.0" encoding="UTF-8"?>
<owms-metadata xmlns:ns0="http://purl.org/dc/terms/" xmlns:ns1="http://standaarden.overheid.nl/oep/meta/" xmlns:ns2="http://standaarden.overheid.nl/owms/terms/">
    <owmskern>
        <ns0:identifier>kst-34200-14</ns0:identifier>
        <ns0:title>Financieel jaarverslag van het Rijk 2014; Motie; Motie van het lid Krol over het continueren van het verlaagde btw-tarief voor verbouwingen en renovaties</ns0:title>
        <ns0:language>nl</ns0:language>
        <ns0:type scheme="OVERHEIDop.KamerstukTypen">Motie</ns0:type>
        <ns0:type scheme="OVERHEIDop.Parlementair">Kamerstuk</ns0:type>
        <ns0:creator scheme="OVERHEID.StatenGeneraal">Tweede Kamer der Staten-Generaal</ns0:creator>
        <dcterms.modified>2015-05-27</dcterms.modified>
    </owmskern>
    <owmsmantel>
        <ns0:available>2015-05-27</ns0:available>
        <ns0:issued>2015-05-26</ns0:issued>
    </owmsmantel>
    <oep>
        <ns1:documentStatus>Opgemaakt na onopgemaakt</ns1:documentStatus>
        <ns1:documenttitel>Motie van het lid Krol over het continueren van het verlaagde btw-tarief voor verbouwingen en renovaties</ns1:documenttitel>
        <ns1:dossiernummer>34200</ns1:dossiernummer>
        <ns1:dossiertitel>Financieel jaarverslag van het Rijk 2014</ns1:dossiertitel>
        <ns1:indiener>H.C.M. Krol</ns1:indiener>
        <ns1:ondernummer>14</ns1:ondernummer>
        <ns1:publicationName>Kamerstuk</ns1:publicationName>
        <ns1:vergaderjaar>2014-2015</ns1:vergaderjaar>
        <ns2:organisationType>staten generaal</ns2:organisationType>
        <ns2:category scheme="OVERHEID.TaxonomieBeleidsagenda">FinanciÃ«n | Begroting</ns2:category>
    </oep>
</owms-metadata>
```

$ curl 'https://zoek.officielebekendmakingen.nl/kst-34200-14/metadata_owms.xml'

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<!--Generatie: versie XSLT 1.1 op 2015-05-27+02:00--><owms-metadata xmlns:dcterms="http://purl.org/dc/terms/"
               xmlns:overheid="http://standaarden.overheid.nl/owms/terms/"
               xmlns:overheidop="http://standaarden.overheid.nl/oep/meta/">
   <owmskern>
      <dcterms:identifier>kst-34200-14</dcterms:identifier>
      <dcterms:title>Financieel jaarverslag van het Rijk 2014; Motie; Motie van het lid Krol over het continueren van het verlaagde btw-tarief voor verbouwingen en renovaties</dcterms:title>
      <dcterms:language>nl</dcterms:language>
      <dcterms:type scheme="OVERHEIDop.KamerstukTypen">Motie</dcterms:type>
      <dcterms:type scheme="OVERHEIDop.Parlementair">Kamerstuk</dcterms:type>
      <dcterms:creator scheme="OVERHEID.StatenGeneraal">Tweede Kamer der Staten-Generaal</dcterms:creator>
      <dcterms.modified>2015-05-27</dcterms.modified>
   </owmskern>
   <owmsmantel>
      <dcterms:available>2015-05-27</dcterms:available>
      <dcterms:issued>2015-05-26</dcterms:issued>
   </owmsmantel>
   <oep>
      <overheidop:documentStatus>Opgemaakt na onopgemaakt</overheidop:documentStatus>
      <overheidop:documenttitel>Motie van het lid Krol over het continueren van het verlaagde btw-tarief voor verbouwingen en renovaties</overheidop:documenttitel>
      <overheidop:dossiernummer>34200</overheidop:dossiernummer>
      <overheidop:dossiertitel>Financieel jaarverslag van het Rijk 2014</overheidop:dossiertitel>
      <overheidop:indiener>H.C.M. Krol</overheidop:indiener>
      <overheidop:ondernummer>14</overheidop:ondernummer>
      <overheidop:publicationName>Kamerstuk</overheidop:publicationName>
      <overheidop:vergaderjaar>2014-2015</overheidop:vergaderjaar>
      <overheid:organisationType>staten generaal</overheid:organisationType>
      <overheid:category scheme="OVERHEID.TaxonomieBeleidsagenda">Financiën | Begroting</overheid:category>
   </oep>
</owms-metadata>
```

### LINKED ISSUE #627: show Garbled message when Content-Type doesn't have charset part
many of website's HTTP header's Content-Type part does not have Charset part, just like:
> Content-Type: text/html

then shows garbled message in terminal/iTerm2 in macOS 10.13  when the html contains Chinese Character which encode with utf8 (which shows OK with curl), but when I change the header like this:
> Content-Type: text/html; charset=utf-8

then OK

the garbled messge is below in  body>h1
```
$ http --debug dev1.cn/test.html
HTTPie 0.9.9
Requests 2.12.3
Pygments 2.1.3
Python 3.6.3 (default, Oct  4 2017, 06:09:15)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)]
/usr/local/Cellar/httpie/0.9.9/libexec/bin/python3.6
Darwin 17.0.0

<Environment {
    "colors": 256,
    "config": {
        "__meta__": {
            "about": "HTTPie configuration file",
            "help": "https://httpie.org/docs#config",
            "httpie": "0.9.9"
        },
        "default_options": "[]"
    },
    "config_dir": "/Users/hh/.httpie",
    "is_windows": false,
    "stderr": "<_io.TextIOWrapper name='<stderr>' mode='w' encoding='UTF-8'>",
    "stderr_isatty": true,
    "stdin": "<_io.TextIOWrapper name='<stdin>' mode='r' encoding='UTF-8'>",
    "stdin_encoding": "UTF-8",
    "stdin_isatty": true,
    "stdout": "<_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>",
    "stdout_encoding": "UTF-8",
    "stdout_isatty": true
}>

>>> requests.request(**{
    "allow_redirects": false,
    "auth": "None",
    "cert": "None",
    "data": {},
    "files": {},
    "headers": {
        "User-Agent": "HTTPie/0.9.9"
    },
    "method": "get",
    "params": {},
    "proxies": {},
    "stream": true,
    "timeout": 30,
    "url": "http://dev1.cn/test.html",
    "verify": true
})

HTTP/1.1 200 OK
Connection: keep-alive
Content-Encoding: gzip
Content-Type: text/html
Date: Sun, 05 Nov 2017 15:23:00 GMT
ETag: W/"59ff2cc9-a7"
Last-Modified: Sun, 05 Nov 2017 15:22:49 GMT
Server: nginx/1.10.3 (Ubuntu)
Transfer-Encoding: chunked

<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8" />
        <title></title>
    </head>
    <body>
        <h1>ä½ å¥½ï¼ä¸ç</h1>
    </body>
</html>
```

the garbled message disturbed me very much, which not exists in curl , but I like httpie more, can you improve it ?

### LINKED ISSUE #1022: Better handling of responses without correct content-type charset
## Problem

If server doesn't provide correct content-type charset, we're defaulting to latin1 [because requests underlying does that](https://github.com/psf/requests/issues/2086). This is undesirable for user experience.

## Possible solutions

* Default to utf8
* For well-known content-types, handle them as defined in respective RFC (utf8 for `application/json`, first line or BOM for XML, meta tag in HTML)
* Use [chardet](https://pypi.org/project/chardet/) to detect from the body

## Considerations
* Streaming mode and chunks
* Discrepancies between streaming and display mode: should this only be done for readability when showing to the terminal, or also when piping to other commands? Current decision: only doing this for terminal display. Consider flag for enforcing this for piping as well
* chardet requires downloading the whole body, so it would download a whole video to tell you it's binary data

BoboTiG:
Supersedes #594.
Fixes #1022 and related already-closed issues.
Fixes #358.
Fixes #627.

### DIFF
--- a/httpie/codec.py
+++ b/httpie/codec.py
@@ -26,3 +26,12 @@ def decode(content: Bytes, encoding: str) -> str:
     if not encoding:
         encoding = detect_encoding(content)
     return content.decode(encoding, 'replace')
+
+
+def encode(content: str, encoding: str) -> bytes:
+    """Encode `content` using the given `encoding`.
+
+    Unicode errors are replaced.
+
+    """
+    return content.encode(encoding, 'replace')
--- a/httpie/codec.py
+++ b/httpie/codec.py
@@ -6,7 +6,7 @@ from .constants import UTF8
 
 
 def detect_encoding(content: bytes) -> str:
-    """Detect the `content` charset encoding.
+    """Detect the `content` encoding.
     Fallback to UTF-8 when no suitable encoding found.
 
     """
--- a/httpie/codec.py
+++ b/httpie/codec.py
@@ -14,7 +14,7 @@ def detect_encoding(content: bytes) -> str:
     return match.encoding if match else UTF8
 
 
-def decode(content: Union[bytearray, bytes, str], encoding: str) -> str:
+def decode(content: Union[bytearray, bytes], encoding: str) -> str:
     """Decode `content` using the given `encoding`.
     If no `encoding` is provided, the best effort is to guess it from `content`.
 
@@ -23,5 +23,4 @@ def decode(content: Union[bytearray, bytes, str], encoding: str) -> str:
     """
     if not encoding:
         encoding = detect_encoding(content)
-
     return content.decode(encoding, 'replace')
--- a/httpie/codec.py
+++ b/httpie/codec.py
@@ -5,7 +5,7 @@ from charset_normalizer import from_bytes
 from .constants import UTF8
 
 
-def detect(content: bytes) -> str:
+def detect_encoding(content: bytes) -> str:
     """Detect the `content` charset encoding.
     Fallback to UTF-8 when no suitable encoding found.
 
@@ -14,24 +14,17 @@ def detect(content: bytes) -> str:
     return match.encoding if match else UTF8
 
 
-class TextDecoderStrategy:
-    """Content decoding strategy.
+def decode(content: Union[bytearray, bytes, str], encoding: str) -> str:
+    """Decode `content` using the provided `encoding`.
+    If no `encoding` is provided, the best effort is to guess it from `content`.
+
+    Unicode errors are replaced.
 
     """
+    if isinstance(content, str):
+        return content
 
-    def __init__(self, encoding: str):
-        self.encoding = encoding
+    if not encoding:
+        encoding = detect_encoding(content)
 
-    def get_encoding(self, content: bytes) -> str:
-        """Return the `content` charset encoding.
-
-        """
-        return self.encoding or detect(content)
-
-    def decode(self, content: Union[bytearray, bytes, str]) -> str:
-        """Decode `content`. Unicode errors are replaced.
-
-        """
-        if isinstance(content, str):
-            return content
-        return content.decode(self.get_encoding(content), 'replace')
+    return content.decode(encoding, 'replace')
--- a/httpie/codec.py
+++ b/httpie/codec.py
@@ -15,15 +15,12 @@ def detect_encoding(content: bytes) -> str:
 
 
 def decode(content: Union[bytearray, bytes, str], encoding: str) -> str:
-    """Decode `content` using the provided `encoding`.
+    """Decode `content` using the given `encoding`.
     If no `encoding` is provided, the best effort is to guess it from `content`.
 
     Unicode errors are replaced.
 
     """
-    if isinstance(content, str):
-        return content
-
     if not encoding:
         encoding = detect_encoding(content)
 
--- a/httpie/codec.py
+++ b/httpie/codec.py
@@ -4,8 +4,10 @@ from charset_normalizer import from_bytes
 
 from .constants import UTF8
 
+Bytes = Union[bytearray, bytes]
 
-def detect_encoding(content: bytes) -> str:
+
+def detect_encoding(content: Bytes) -> str:
     """Detect the `content` encoding.
     Fallback to UTF-8 when no suitable encoding found.
 
@@ -14,7 +16,7 @@ def detect_encoding(content: bytes) -> str:
     return match.encoding if match else UTF8
 
 
-def decode(content: Union[bytearray, bytes], encoding: str) -> str:
+def decode(content: Bytes, encoding: str) -> str:
     """Decode `content` using the given `encoding`.
     If no `encoding` is provided, the best effort is to guess it from `content`.
 
--- a/httpie/codec.py
+++ b/httpie/codec.py
@@ -1,0 +1,37 @@
+from typing import Union
+
+from charset_normalizer import from_bytes
+
+from .constants import UTF8
+
+
+def detect(content: bytes) -> str:
+    """Detect the `content` charset encoding.
+    Fallback to UTF-8 when no suitable encoding found.
+
+    """
+    match = from_bytes(bytes(content)).best()
+    return match.encoding if match else UTF8
+
+
+class TextDecoderStrategy:
+    """Content decoding strategy.
+
+    """
+
+    def __init__(self, encoding: str):
+        self.encoding = encoding
+
+    def get_encoding(self, content: bytes) -> str:
+        """Return the `content` charset encoding.
+
+        """
+        return self.encoding or detect(content)
+
+    def decode(self, content: Union[bytearray, bytes, str]) -> str:
+        """Decode `content`. Unicode errors are replaced.
+
+        """
+        if isinstance(content, str):
+            return content
+        return content.decode(self.get_encoding(content), 'replace')
--- a/httpie/models.py
+++ b/httpie/models.py
@@ -31,11 +31,6 @@ class HTTPMessage(metaclass=ABCMeta):
         """Return a `str` with the message's encoding, if known."""
 
     @property
-    def body(self) -> bytes:
-        """Return a `bytes` with the message's body."""
-        raise NotImplementedError()
-
-    @property
     def content_type(self) -> str:
         """Return the message content type."""
         ct = self._orig.headers.get('Content-Type', '')
@@ -85,12 +80,6 @@ class HTTPResponse(HTTPMessage):
     @property
     def encoding(self):
         return self._orig.encoding or UTF8
-
-    @property
-    def body(self):
-        # Only now the response body is fetched.
-        # Shouldn't be touched unless the body is actually needed.
-        return self._orig.content
 
 
 class HTTPRequest(HTTPMessage):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for formatting & coloring of JSON bodies preceded by non-JSON data (e.g., an XXSI prefix). ([#1130](https://github.com/httpie/httpie/issues/1130))
 - Added `--format-options=response.as:CONTENT_TYPE` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
 - Added `--response-as` shortcut for setting the response `Content-Type`-related `--format-options`. ([#1134](https://github.com/httpie/httpie/issues/1134))
+- Improved handling of responses without correct `Content-Type` charset. ([#1110](https://github.com/httpie/httpie/issues/1110))
 - Installed plugins are now listed in `--debug` output. ([#1165](https://github.com/httpie/httpie/issues/1165))
 - Fixed duplicate keys preservation of JSON data. ([#1163](https://github.com/httpie/httpie/issues/1163))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for formatting & coloring of JSON bodies preceded by non-JSON data (e.g., an XXSI prefix). ([#1130](https://github.com/httpie/httpie/issues/1130))
 - Added `--format-options=response.as:CONTENT_TYPE` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
 - Added `--response-as` shortcut for setting the response `Content-Type`-related `--format-options`. ([#1134](https://github.com/httpie/httpie/issues/1134))
-- Improved handling of responses without correct `Content-Type` charset. ([#1110](https://github.com/httpie/httpie/issues/1110))
+- Improved handling of responses without correct `Content-Type` encoding. ([#1110](https://github.com/httpie/httpie/issues/1110))
 - Installed plugins are now listed in `--debug` output. ([#1165](https://github.com/httpie/httpie/issues/1165))
 - Fixed duplicate keys preservation of JSON data. ([#1163](https://github.com/httpie/httpie/issues/1163))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for formatting & coloring of JSON bodies preceded by non-JSON data (e.g., an XXSI prefix). ([#1130](https://github.com/httpie/httpie/issues/1130))
 - Added `--format-options=response.as:CONTENT_TYPE` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
 - Added `--response-as` shortcut for setting the response `Content-Type`-related `--format-options`. ([#1134](https://github.com/httpie/httpie/issues/1134))
-- Improved handling of responses without correct `Content-Type` encoding. ([#1110](https://github.com/httpie/httpie/issues/1110))
+- Improved handling of prettified responses without correct `Content-Type` encoding. ([#1110](https://github.com/httpie/httpie/issues/1110))
 - Installed plugins are now listed in `--debug` output. ([#1165](https://github.com/httpie/httpie/issues/1165))
 - Fixed duplicate keys preservation of JSON data. ([#1163](https://github.com/httpie/httpie/issues/1163))
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1249,6 +1249,18 @@ For example, the following request will force the response to be treated as XML:
 $ http --response-as=application/xml pie.dev/get
 ```
 
+And the following requests will force the response to use the charset encoding [big5](https://docs.python.org/3/library/codecs.html#standard-encodings):
+
+```bash
+$ http --response-as='charset=big5' pie.dev/get
+```
+
+```bash
+$ http --response-as='text/plain; charset=big5' pie.dev/get
+```
+
+Given the charset encoding is not sent by the server, HTTPie will auto-detect it.
+
 ### Binary data
 
 Binary data is suppressed for terminal output, which makes it safe to perform requests to URLs that send back binary data.
--- a/docs/README.md
+++ b/docs/README.md
@@ -1259,7 +1259,7 @@ $ http --response-as='charset=big5' pie.dev/get
 $ http --response-as='text/plain; charset=big5' pie.dev/get
 ```
 
-Given the charset encoding is not sent by the server, HTTPie will auto-detect it.
+Given the encoding is not sent by the server, HTTPie will auto-detect it.
 
 ### Binary data
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1249,7 +1249,7 @@ For example, the following request will force the response to be treated as XML:
 $ http --response-as=application/xml pie.dev/get
 ```
 
-And the following requests will force the response to use the charset encoding [big5](https://docs.python.org/3/library/codecs.html#standard-encodings):
+And the following requests will force the response to use the [big5](https://docs.python.org/3/library/codecs.html#standard-encodings) encoding:
 
 ```bash
 $ http --response-as='charset=big5' pie.dev/get
--- a/httpie/cli/definition.py
+++ b/httpie/cli/definition.py
@@ -316,6 +316,7 @@ output_processing.add_argument(
     Override the response Content-Type for formatting purposes, e.g.:
 
         --response-as=application/xml
+        --response-as='application/xml; charset=utf-8'
 
     It is a shortcut for:
 
--- a/httpie/cli/definition.py
+++ b/httpie/cli/definition.py
@@ -316,6 +316,7 @@ output_processing.add_argument(
     Override the response Content-Type for formatting purposes, e.g.:
 
         --response-as=application/xml
+        --response-as=charset=utf-8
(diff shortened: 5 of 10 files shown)


BoboTiG:
`packit` failures are OK: the new test dependency is not yet known there.

tests/test_httpie.py
```diff
@@ -218,3 +219,21 @@ def test_json_input_preserve_order(httpbin_both):
     assert HTTP_OK in r
     assert r.json['data'] == \
         '{"order": {"map": {"1": "first", "2": "second"}}}'
+
+
+@responses.activate
+def test_GET_encoding_detection_from_content():
+    url = 'http://example.org'  # Note: URL never fetched
+    body = '<?xml version="1.0" encoding="UTF-8"?><category>Financiën</category>'.encode('utf8')
+    responses.add(responses.GET, url, body=body, content_type='text/xml')
```
BoboTiG, line 228: I am not sure if `httpbin` does allow doing such things. @jakubroztocil Let me know if you think of a way of doing that without a new dependency (I am OK with that, but if we can keep requirements light, it is always good).
And in the other way I really like `responses`, it is very easy to mock calls without touching `requests` internals ourselves.

BoboTiG:
I would like to change one thing or two. No need to review it for now.

jakubroztocil (commented):
How is this different from accessing `response.apparent_encoding`? The main problem with `apparent_encoding` is that it breaks our output streaming behavior.   

How are we making sure the response body is still processed in chunks as opposed to being buffered first?

BoboTiG:
> How is this different from accessing `response.apparent_encoding`? The main problem with `apparent_encoding` is that it breaks our output streaming behavior.
>
> How are we making sure the response body is still processed in chunks as opposed to being buffered first?

The workaround/solution is done in 2 times:
1. We do not need to get the response encoding to display headers. The old way was a blocker: to display headers, we would fetch the body to get its encoding. Now, as headers *should* contain only US-ASCII characters, we use the terminal encoding or fallback to UTF-8 (it seems the more robust idea).
2. It seems we never need to fetch the whole body directly, we either use `HTTPResponse.iter_body()` either `HTTPResponse.iter_lines()`. I reworked both to store the first chunk into `HTTPResponse._body` and use it to guess the whole body encoding. Why does it work? Because we need to encode stuff **after** having retrieved the first chunk, so we can properly guess the encoding based on that first chunk.

The only potential problem I could see with part 2 is that the guessed encoding is not the good one (because there is no enough data to properly guess the right encoding). I think we are quite good with the new implementation and I do not have a solution for such hypothetical problem for now.

I've had hard time adding pure stream tests. So I simply added tests to cover reported problems, and the covered code is still good (I did not introduced uncovered code).

BoboTiG:
I just found a regression:
```shell
$ python -m httpie --pretty=all --stream pie.dev/stream/3  | while read line; do echo "$line" ; done
{
"args": {},
"headers": {
"Accept": "*/*",
"Accept-Encoding": "gzip",
"Cdn-Loop": "cloudflare",
"Cf-Connecting-Ip": "89.159.87.157",
"Cf-Ipcountry": "FR",
"Cf-Ray": "676d5d1fa80ecdc3-FRA",
"Cf-Visitor": "{"scheme":"http"}",
"Connection": "Keep-Alive",
"Host": "pie.dev",
"User-Agent": "HTTPie/2.5.0.dev0"
},
"id": 0,
"origin": "89.159.87.157",
"ur
```
2 calls are misisng. I'm investigating.

BoboTiG:
Fixed.

I've another concern about untested code (streaming with converters). I'll add a non-regression test.

tests/test_httpie.py
BoboTiG: I'll see if I can get rid of `responses`, I'm not comfortable to add a `requests`-specific module to mock calls. Trying to stick with `httpbin` would be better.

BoboTiG:
> I've another concern about untested code (streaming with converters). I'll add a non-regression test.

Test added, my concern was not justified :)

tests/test_httpie.py
BoboTiG: Finally, it is not worth losing time on it. I did not find how to do it and at the end `httpbin` is also tight to `resquests`.

BoboTiG:
@jakubroztocil we are good for a review. If you prefer, I can open a specific PR for commits 1 and 2 as it would ease the review. LMK.

tests/test_httpie.py
```diff
@@ -218,3 +219,21 @@ def test_json_input_preserve_order(httpbin_both):
...
+    assert 'Financiën' in r
+
+
+@responses.activate
+def test_POST_encoding_detection_from_content():
+    url = 'http://example.org'  # Note: URL never fetched
+    body = 'Všichni lidé jsou si rovni.'.encode('utf8')
+    responses.add(responses.POST, url, body=body, content_type='text/plain')
+    r = http('--form', 'POST', url)
+    assert 'Všichni lidé jsou si rovni.' in r
```
jakubroztocil, line 239: Since the responses use `utf8`, and the implementation now also uses `utf8` as a fallback, this doesn’t actually test the content encoding detection. It’d also need to be tested with non-`utf8` encodings. And when these responses are encoded as `windows-1250`, for example, then the tests fail.

jakubroztocil:
Does this ignore the `Content-Type` charset when specified? I believe the auto-detection should only kick in when the message doesn’t say. 

I’m also afraid that relying on the first chunk only (i.e., the first line) won’t be very reliable. I’d imagine applying some heuristic to each chunk until we build confidence (e.g., look for a non-whitespace / non-ASCII chunk). 

This also fails:

```python
# windows-1250 response with a correctly specified charset
@responses.activate
def test_POST_encoding_detection_from_content():
    url = 'http://example.org'  # Note: URL never fetched
    body = 'Všichni lidé jsou si rovni.'.encode('windows-1250')
    responses.add(responses.POST, url, body=body, content_type='text/plain; charset=windows-1250')
    r = http('--form', 'POST', url)
    assert 'Všichni lidé jsou si rovni.' in r
```

The first line can also be something like `<?xml version="1.0" encoding="windows-1250"?>`, for example, which the current implementation would detect as UTF-8 (even with a properly set `charset` in `Content-Type`).

Ousret:
Hi there,

Clearly, I must agree with @jakubroztocil doing the assumption that the first chunk would give us an accurate guess is not the way to go.
Imagine the first line being `<?xml version="1.0">`, how would you guess something else than ASCII?

Secondly, fallback to UTF-8 for everything is quite dangerous for reliability. What I would suggest is to try out the detection then use the UTF-8 decoder using `error="strict"` then latin_1 with `error="replace"` if the second and first would fail.

I would not recommend building confidence yourself using chunks but rather use as many bytes as possible, performances should not be a concern most of the time.

Happy to help if I can do anything to ease your task up ahead.

BoboTiG:
> I would not recommend building confidence yourself using chunks but rather use as many bytes as possible, performances should not be a concern most of the time. If you can use the whole body, do it to assure you the best possible guess.

We do not want to use the wole body, that would kill the streaming stuff.

A concern I had was: OK, let's guess the encoding from the chunk 1. It return UTF-8. If I try to decode stuff using more data (more chunks), and it the detected encoding changes, that will be a mess to handle.

I'm not sure how to warkaround that.

What do you have in mind when you say "use as many bytes as possible"? I mean, the encoding is needed right after the first chunk, I do not see yet how to get more data without storing the full body (and losing the streaming stuff). Even "part of the body" is subjective, maybe fetching and storing first N chunks would lower the problem.

I completely agree that the current proposed approach is too naive, let's improve it :)

BoboTiG:
> Does this ignore the `Content-Type` `charset` when specified? I believe the auto-detection should only kick in when the message doesn’t say.

Indeed. I'll fix it.

Ousret:
> I do not see yet how to get more data without storing the full body (and losing the streaming stuff). 

Me neither, I have not studied the core code enough. But I am starting to get a better understanding.

> What do you have in mind when you say "use as many bytes as possible"?

What I had in mind, was to store the whole body in case httpie would need to establish a fair guess of what the encoding actually is.  And I think we all agree upon this.

This seems like a challenge, and I like challenges. How to establish a fair guess with the whole content without breaking the stream I/O? 

The fact that we would need the guess at the first read-->stdout would require temporizing the content in **that specific case**, then rebuilding the stream using **utf-8** bytes maybe?.

I will think more about it.

tests/test_httpie.py
BoboTiG: Fixed :heavy_check_mark:

BoboTiG:
I pushed a patch to take into account the provided charset first, and then fallback to detection if none is specified or if the encoding detected by `requests` is empty.

BoboTiG:
I added streaming tests. And as you both pointed it `test_encoding_detection_from_content_with_no_declared_encoding_at_all()` is failing because for now we are relying on the first chunk of data to guess the encoding: `ascii` is found and it does not work.

BoboTiG:
02b695f32aca0e9f98941d62918c7f1bbe40c141 improves encoding detection by looking at 5 chunks instead of the 1st one. It is a simple commit to demonstrate the thing.
It may be simple and naive, but WDYT about that? It may be enough to workaround most of use cases :thinking:

Ousret:
Hi @BoboTiG 

For me `iter_content` based on a **line** (eg. LF) is unpredictable. Impossible to anticipate the bytes count that will be fed into the detector. Five chunks are going to be better than a single for sure, but I am somewhat displeased by it. If someone told me, how many bytes should I put into the detector, I would answer 512 bytes would be the minimum, 2k would be nice, the whole thing should be perfect.
Still, I am curious about what kind of results/coverages you may get. You could demonstrate the reliability using something like:

- https://github.com/Ousret/charset_normalizer/blob/master/bin/serve.py
- https://github.com/Ousret/charset_normalizer/blob/master/.github/workflows/integration.yml

I really fouth about it, and there is no clean way to do it without disturbing the actual flow. For me, the missing charset should be treated as an exceptional case (eg. outside the normal streaming flow).

BoboTiG:
I reworked the code to consume at least 512 bytes (not yet pushed). I'll integrate your test files to demonstrate (or not) the robustness of the new implementation. Thanks for your inputs, very valuable :)

httpie/models.py
```diff
@@ -45,11 +46,44 @@ def content_type(self) -> str:
...
+        Return the chunk for convenience.
+
+        """
+        try:
+            self._first_chunk_body = []
+            for _ in range(5):
+                self._first_chunk_body.append(next(it))
+        except StopIteration:
+            pass
+        return self._first_chunk_body
```
jakubroztocil, line 72: this breaks streaming

BoboTiG:
@Ousret I push my changes. I am not sure new tests are good.

Have a look at 7dd8b6354a02a06f002522966e3bdebdc8286214, and the new test `test_encoding_detection_from_content_with_no_declared_encoding_at_all`. It uses `from_bytes(file.read_bytes()).best().encoding` to compare what HTTPie found with what `charset_normalizer` would fine using th eentire file. But from what I see, results are not good (sometimes the encoding is `None`. Could you tell me what I've done wrong :pray:?

httpie/models.py
BoboTiG: I know and I am trying different approaches, nothing really working for now.
Also I reviewed that part with d826da219acd92f0f5b18b28675f1f2ef221fbcc to not break streaming when the encoding was effectively already detected. That lowers a little bit the impact but still break streaming for other use cases.

Ousret:
@BoboTiG I can't take a deeper look right now, but according to https://github.com/Ousret/charset_normalizer/runs/3220902913 you should get about three files that do not resolve to anything. Can't get perfect coverage.

jakubroztocil:
I recommend taking a step back and really thinking about the problem, the use cases, and the different streaming-related modes in which HTTPie operates.

**Q: What problem are we trying to solve?**
A: Poor user experience when a response doesn’t specify `charset,` and the output gets misencoded.

**Q: How can we solve the problem?**
A: Use a charset detection library as a fallback.

**Q: What are the constraints?**
A: HTTPie uses a streaming model, which is incompatible with robust charset detection. We also don’t alter data unless the user asks for it.

**Q: Okay, but are there any scenarios in which HTTPie fully loads a response before outputting it and altering data is okay?**
A: Yes. When “prettifying” a response, it’s buffered first. Also, with `--stream`, each line can be considered a standalone document.

**Q: Right, so what if we focus only on the buffered and `--stream` modes?**
A: Yes, in fact, that would capture the majority of use cases. One can also argue that charset detection naturally fits the prettification concept.

**Q: Okay, if we go with this, what would the practical implications be?**
A: When a user makes a simple request on the terminal (`http pie.dev/get`), the charset detection fallback would be used (by default, we prettify, therefore buffer). Same with `--stream`. If the user disables prettifying (`http --pretty=none pie.dev/get`), we stream, so no auto-detection is possible (i.e., no change in behavior here).

**Q: Is there anything else that should be considered?**
A: Yes, this change alters a core behavior, so it needs to be approached holistically and think about all the possible scenarios. This QA is not comprehensive. The behavior also needs to be documented (a new section on charset handling would be a good idea).

BoboTiG:
> **Q: Okay, but are there any scenarios in which HTTPie fully loads a response before outputting it and altering data is okay?**
> A: Yes. When “prettifying” a response, it’s buffered first. Also, with `--stream`, each line can be considered a standalone document.

Considering each line a standalone document is an idea. I am afraid that can lead to exotic outputs (line 1 is detected as ASCII, line 2 as CP1250, line N as Johab, etc).

> **Q: Okay, if we go with this, what would the practical implications be?**
> A: When a user makes a simple request on the terminal (`http pie.dev/get`), the charset detection fallback would be used (by default, we prettify, therefore buffer). Same with `--stream`. If the user disables prettifying (`http --pretty=none pie.dev/get`), we stream, so no auto-detection is possible (i.e., no change in behavior here).

`http pie.dev/get` will use the `EncodedStream` class. The body is always streamed by line, which leads to the same issues as for `--stream`. It is the same scenario if we pass `--pretty=none`. That is because each line needs to be decoded before being buffered.

`--pretty=all|colors|format` will use the `BufferedPrettyStream` class, here as your said: body is buffered.

BoboTiG:
[Still WIP]

The current implementation (6bbfb498cfcce3cc7c1df5386bece9ce304c8328) will store the body retrieved so far to detect the encoding. Results should be more and more accurate with the amount of data. But this leads to potential memory issues and exotic outputs (line 1 would be detected as UTF-8 instead of Johab, line 2 as ASCII, etc.). At the end, it would be better to assume UTF-8 (or something else) for all lines instead of mixing encodings.

And the encoding detection is done for each chunks (for now) which may impact performances.

BoboTiG:
Side note: Fedora 33 and 34 supports will not be lost due to the new requirement (`charset-normalizer`, see packaging failures). It  will be good when @hroncok will have time to backport the project to those versions (thanks a lot!).

hroncok:
Successfully waited for python-charset-normalizer-2.0.4-1.fc33 to appear in the f33-build repo
Successfully waited for python-charset-normalizer-2.0.4-1.fc34 to appear in the f34-build repo

If you include https://github.com/httpie/httpie/pull/1119, you should be able to run the tests on Fedora 33 and 34.

hroncok:
The current Fedora problem is:

```
No matching package to install: 'python3dist(requests) >= 2.26'
No matching package to install: 'python3dist(requests[socks]) >= 2.26'
```

BoboTiG:
Yes I was trying if it was working with that new version. I'll revert that change.

BoboTiG:
So I adapted the current implementation:

- When the `Content-Type` header is set: no behavior change.

- For streamed responses, the encoding is guessed from the last chunk retrieved (best effort). That means the encoding can be different given the line that is streamed. Each line/chunk is treated as a standalone snippet.

jakubroztocil:
> `http pie.dev/get` will use the `EncodedStream` class. The body is always streamed by line, which leads to the same issues as for `--stream`. It is the same scenario if we pass `--pretty=none`. That is because each line needs to be decoded before being buffered.
> `--pretty=all|colors|format` will use the BufferedPrettyStream class, here as your said: body is buffered.


It uses `BufferedPrettyStream` because `--pretty=all` is the default. https://httpie.io/docs#terminal-output

httpie/codec.py
```diff
@@ -0,0 +1,77 @@
...
+    Fallback to UTF-8 when no suitable encoding found.
+
+    """
+    from charset_normalizer import from_bytes
+
+    match = from_bytes(bytes(content)).best()
+    return match.encoding if match else UTF8
+
+
+class Codec:
```
Ousret, line 22: In my opinion, it is a bit misleading. `Codec` does not reflect what you intend to do here.
Perhaps `TextDecoderStrategy` instead?

httpie/codec.py
```diff
@@ -0,0 +1,77 @@
...
+from .constants import UTF8
+from .context import Environment
+
+
+def detect(content: bytes) -> str:
+    """Detect the charset based on `content`.
+    Fallback to UTF-8 when no suitable encoding found.
+
+    """
+    from charset_normalizer import from_bytes
```
Ousret, line 16: Why making the import here? I am curious.

BoboTiG: At first it was to lazy import the module. But now that I thought about it, `request` 2.26.0+ will already import it. I(ll move it to the top, thank you for the catch.

httpie/codec.py
BoboTiG: Indeed, it seems more appropriate :+1:

PR #1111, opened 2021-07-20
Simplify spinner_pos calculation a little

ainzzorl:
A teeny-tiny refactoring. I noticed this condition while skimming the code and though it looked awkward :-)


jakubroztocil:
This is cleaner and more elegant indeed. Thanks, @ainzzorl!

PR #1112, opened 2021-07-23
Minor clean-up

BoboTiG:
I had a look at tests coverage in the context of #1110 and saw 1-2 things that could be quickly tackled.

### DIFF
--- a/httpie/context.py
+++ b/httpie/context.py
@@ -119,10 +119,6 @@ class Environment:
             self._devnull = open(os.devnull, 'w+')
         return self._devnull
 
-    @devnull.setter
-    def devnull(self, value):
-        self._devnull = value
-
     def log_error(self, msg, level='error'):
         assert level in ['error', 'warning']
         self._orig_stderr.write(f'\n{self.program_name}: {level}: {msg}\n\n')
--- a/httpie/downloads.py
+++ b/httpie/downloads.py
@@ -2,7 +2,6 @@
 Download mode implementation.
 
 """
-import errno
 import mimetypes
 import os
 import re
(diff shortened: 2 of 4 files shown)


httpie/core.py
```diff
@@ -50,10 +50,8 @@ def main(args: List[Union[str, bytes]] = sys.argv, env=Environment()) -> ExitSta
     exit_status = ExitStatus.SUCCESS
 
     try:
-        parsed_args = parser.parse_args(
-            args=args,
-            env=env,
-        )
+        parsed_args = parser.parse_args(args=args, env=env)
+        exit_status = program(args=parsed_args, env=env)
```
jakubroztocil, line 54: `parse_args()` and `program()` each have their own `try` clause intentionally because we handle different types of exceptions for them. For example, we don’t handle base `Exception`s from `parse_args()` as you can see. We also want to keep the `try` branches minimal and the distinction explicit. To get rid of the duplicate handling of `SystemError` and `KeyboardInterrupt`, something like this would be more in line with the intended behavior:

```python
try:
    parse_args()
    try:
        program()
    except Timeout:
        pass
    except TooManyRedirects:
        pass
    except Exception:
        pass
except KeyboardInterrupt:
    pass
except SystemExit:
    pass
```

jakubroztocil (changes requested):

httpie/core.py
jakubroztocil: This approach would come with the implicit requirement that the outer `except` types don’t inherit from `Exception`, though.

BoboTiG: Good catch. I removed the commit as it is not really interesting to change that part for now.

PR #1113, opened 2021-07-26
--download: Use `time.monotonic()` and rework code to prevent `ZeroDivisionError` specific handling

BoboTiG:
It simplifies speed/percentage handling. It also expands coverage as testing previous `ZeroDivisionError` corner-cases is not easy.

### DIFF
--- a/httpie/downloads.py
+++ b/httpie/downloads.py
@@ -9,7 +9,7 @@ import re
 import sys
 import threading
 from mailbox import Message
-from time import sleep, time
+from time import sleep, monotonic
 from typing import IO, Optional, Tuple
 from urllib.parse import urlsplit
 
@@ -350,7 +350,7 @@ class DownloadStatus:
         assert self.time_started is None
         self.total_size = total_size
         self.downloaded = self.resumed_from = resumed_from
-        self.time_started = time()
+        self.time_started = monotonic()
 
     def chunk_downloaded(self, size):
         assert self.time_finished is None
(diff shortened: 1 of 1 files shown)


httpie/downloads.py
```diff
@@ -22,6 +22,8 @@
 
 PARTIAL_CONTENT = 206
 
+TIME_FUNC = monotonic
```
jakubroztocil, line 25: Why the constant?

jakubroztocil (changes requested):
Using `monotonic` is a good idea. However, it doesn’t remove the need for the handling of `ZeroDivisionError`.

httpie/downloads.py
```diff
@@ -407,26 +409,22 @@ def run(self):
...
 
         if now - self._prev_time >= self._update_interval:
             downloaded = self.status.downloaded
-            try:
-                speed = ((downloaded - self._prev_bytes)
-                         / (now - self._prev_time))
-            except ZeroDivisionError:
-                speed = 0
+            speed = ((downloaded - self._prev_bytes)
+                     / (now - self._prev_time))
```
jakubroztocil, line 417: `downloaded - self._prev_bytes` can be `0` if nothing new has been downloaded

BoboTiG: Yes, and it would not end on a `ZeroDivisionError`, it is safe.

httpie/downloads.py
BoboTiG: To ease monkey patching or eventual future change. I will revert, it is not really necessary.

BoboTiG:
The title is misleading, you are right. I'll reword it.

PR #1115, opened 2021-07-28
Uniformize UTF-8 naming

BoboTiG:
Replace `utf8` -> `utf-8` everywhere.
It should have no impact, `utf8` is an alias of `utf-8` [1].

[1] https://github.com/python/cpython/blob/ee03bad25e83b00ba5fc2a0265b48c6286e6b3f7/Lib/encodings/aliases.py#L534

---

Full story: I kind of need such uniformization for #1110 where I am adding tests. And I have to deal with our naming (`utf8`) and the one from `requests` or `charset_normalizer` (`utf-8`).
Note that this is a one-time small clean-up, and I can live without that patch being merged.

### DIFF
--- a/httpie/constants.py
+++ b/httpie/constants.py
@@ -1,0 +1,2 @@
+# UTF-8 encoding name
+UTF8 = 'utf-8'
--- a/httpie/constants.py
+++ b/httpie/constants.py
@@ -1,2 +1,2 @@
 # UTF-8 encoding name
-UTF8 = 'utf-8'
+UTF8 = 'utf_8'
--- a/httpie/constants.py
+++ b/httpie/constants.py
@@ -1,2 +1,2 @@
 # UTF-8 encoding name
-UTF8 = 'utf_8'
+UTF8 = 'utf-8'
--- a/httpie/cli/argparser.py
+++ b/httpie/cli/argparser.py
@@ -24,6 +24,7 @@ from .constants import (
 from .exceptions import ParseError
 from .requestitems import RequestItems
 from ..context import Environment
+from ..constants import UTF8
 from ..plugins.registry import plugin_manager
 from ..utils import ExplicitNullAuth, get_content_type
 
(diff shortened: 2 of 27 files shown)


BoboTiG:
It is finally not needed for #1110 but I am still for such clean-up, if you are OK with that @jakubroztocil :)

jakubroztocil (commented):
We can remove the argument from `.encode('utf-8')` and `.decode('utf-8')` since it’s the default. 

I’d also introduce a `UTF8 = 'UTF-8'` constant and use it instead of the string to avoid inconsistencies and to have a central place where we can about the encoding.

BoboTiG:
Actually I did the exact opposite :)

I think it is a good thing to explicitly set the encoding everywhere to prevent surprises. And the [PEP-597](https://www.python.org/dev/peps/pep-0597/#using-the-default-encoding-is-a-common-mistake) will eventually throw warnings in a near future.

httpie/constants.py
```diff
@@ -0,0 +1,2 @@
+# UTF-8 encoding name
+UTF8 = 'utf-8'
```
BoboTiG, line 2:
```suggestion
UTF8 = 'utf_8'
```

jakubroztocil:
> Actually I did the exact opposite :)
> 
> I think it is a good thing to explicitly set the encoding everywhere to prevent surprises. And the [PEP-597](https://www.python.org/dev/peps/pep-0597/#using-the-default-encoding-is-a-common-mistake) will eventually throw warnings in a near future.

I believe the PEP only talks about `open()`’s encoding. For `encode()`/`decode()`, `utf-8` is the argument’s default value:


```python
str.encode(encoding="utf-8", errors="strict")
bytes.decode(encoding="utf-8", errors="strict")¶
```

https://docs.python.org/3/library/stdtypes.html#str.encode
https://docs.python.org/3/library/stdtypes.html#bytes.decode

BoboTiG:
Oh how did miss that?!

I'll remove all encoding occurrences for those functions then 👍

PR #1117, opened 2021-08-04
Add more streaming tests

BoboTiG:
I extracted several tests from #1110 as they are not directly related to that work and it will ease reviewing the former.

Also fixed minor glitches here and there and re-enabled a unicode test.

### DIFF
--- a/tests/test_stream.py
+++ b/tests/test_stream.py
@@ -15,7 +15,7 @@ from .fixtures import BIN_FILE_CONTENT, BIN_FILE_PATH
 PRETTY_OPTIONS = list(PRETTY_MAP.keys())
 
 
-class FakeConverterPlugin(ConverterPlugin):
+class SortJSONConverterPlugin(ConverterPlugin):
     @classmethod
     def supports(cls, mime):
         return mime == 'json/bytes'
@@ -59,9 +59,10 @@ def test_pretty_stream_ensure_full_stream_is_retrieved(httpbin):
 @pytest.mark.parametrize('stream', [True, False])
 @responses.activate
 def test_pretty_options_with_and_without_stream_with_converter(pretty, stream):
-    plugin_manager.register(FakeConverterPlugin)
+    plugin_manager.register(SortJSONConverterPlugin)
     try:
-        assert 'FakeConverterPlugin' in str(plugin_manager)
+        # Cover PluginManager.__repr__()
+        assert 'SortJSONConverterPlugin' in str(plugin_manager)
 
         url = 'http://example.org'  # Note: URL never fetched
         body = b'\x00{"foo":42,\n"bar":"baz"}'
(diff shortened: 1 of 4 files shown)


tests/test_stream.py
```diff
@@ -24,6 +44,45 @@ def test_pretty_redirected_stream(httpbin):
...
+    assert r.count('/stream/3') == 3
+
+
+@pytest.mark.parametrize('pretty', PRETTY_OPTIONS)
+@pytest.mark.parametrize('stream', [True, False])
+@responses.activate
+def test_pretty_options_with_and_without_stream_with_converter(pretty, stream):
+    plugin_manager.register(FakeConverterPlugin)
+    try:
+        assert 'FakeConverterPlugin' in str(plugin_manager)
```
jakubroztocil, line 64: Why the string search instead of just `assert FakeConverterPlugin in plugin_manager`?

tests/test_stream.py
```diff
@@ -24,6 +44,45 @@ def test_pretty_redirected_stream(httpbin):
...
+        if stream:
+            args.insert(0, '--stream')
+        r = http(*args)
+
+        assert 'json/bytes' in r
+        if pretty == 'none':
+            assert BINARY_SUPPRESSED_NOTICE.decode() in r
+        else:
+            assert '"bar": "baz",' in r
+            assert '"foo": 42' in r
```
jakubroztocil, line 81: The JSON object [should be accessible](https://github.com/httpie/httpie/blob/350f973f705dc996af27d66fe02e03e2f1f99968/tests/utils/__init__.py#L139-L171) via `r.json`. Is it not working in this case? If it is, then this is preferable:

```python
assert r.json['bar'] == 'baz'
```

tests/test_stream.py
```diff
@@ -1,10 +1,30 @@
...
+from httpie.plugins import ConverterPlugin
+from httpie.plugins.registry import plugin_manager
+
 from .utils import StdinBytesIO, http, MockEnvironment
 from .fixtures import BIN_FILE_CONTENT, BIN_FILE_PATH
 
+PRETTY_OPTIONS = list(PRETTY_MAP.keys())
+
+
+class FakeConverterPlugin(ConverterPlugin):
```
jakubroztocil, line 18: It’s not _completely_ fake. It does something so I’d vote for giving it a slightly more meaningful name.

jakubroztocil (changes requested):

tests/test_stream.py
BoboTiG: It is to cover `PluginsManager.\_\_repr\_\_()`.

tests/test_stream.py
jakubroztocil: I see! This comment would make a useful `# comment`.

PR #1127, opened 2021-08-13
Fix handling of session files with `Cookie:` followed by other headers

### LINKED ISSUE #1126: Session file with cookie cannot be parsed
**Checklist**

- [Y] I've searched for similar issues.
- [Y] I'm using the the latest version of HTTPie.

- httpie version, 2.4.0
- python version, 3.9

I prepared the **my-session-cookie.json** file, the request parameters are placed in **query.json**, and the expected response results are placed in the **result.json** file.

 **my-session-cookie.json** like this:
`
{
      "__meta__": {
            "about": "HTTPie session file",
            "help": "https://httpie.org/doc#sessions",
            "httpie": "2.4.0"
       },
       "headers": {
            "Content-Type": "application/json",
            "cookie": "12345"
      }
}
`

And command like this:

Xnip2021-08-13_02-22-22

`http --verify=no -v  --session-read-only=~/Desktop/my-session-cookie.json POST http://localhost:8301/test < ~/Desktop/query.json -d >>~/Desktop/result.json`

And I got error like this:
`http: error: RuntimeError: OrderedDict mutated during iteration`

Then I did some test and the request can be executed
- remove cookie field
- OR remove `-d >>~/Desktop/result.json` in the command

So what's wrong？

BoboTiG:
Fixes #1126.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
   an alternative to ``stdin``. (`#534`_)
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
 * Fixed ``--verbose`` HTTP 307 redirects with streamed request body. (`#1088`_)
+* Fixed session files handling when there are cookies and other headers following. (`#1126`_)
 * Add internal support for file-like object responses to improve adapter plugin support. (`#1094`_)
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
   an alternative to ``stdin``. (`#534`_)
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
 * Fixed ``--verbose`` HTTP 307 redirects with streamed request body. (`#1088`_)
-* Fixed session files handling when there are cookies and other headers following. (`#1126`_)
+* Fixed handling of session files with `Cookie:` followed by other headers. (`#1126`_)
 * Add internal support for file-like object responses to improve adapter plugin support. (`#1094`_)
 
 
--- a/tests/test_sessions.py
+++ b/tests/test_sessions.py
@@ -179,9 +179,9 @@ class TestSession(SessionTestBase):
         assert r2.json['headers']['Foo'] == 'Bar'
 
     def test_session_with_cookie_followed_by_another_header(self, httpbin):
-        """Non-regression test for issue #1126 where request headers were mutated during iteration:
-
-            RuntimeError: OrderedDict mutated during iteration
+        """
+        Make sure headers don’t get mutated — <https://github.com/httpie/httpie/issues/1126>
+        
         """
         self.start_session(httpbin)
         session_data = {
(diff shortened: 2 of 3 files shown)


BoboTiG:
The exact issue happens when there is another header to handled after we do the `del request_headers[name]`.

Here is a reproduction case to ease understanding:
```python
# file: test.py
from requests.structures import CaseInsensitiveDict

d = CaseInsensitiveDict({
    "a": 1,
    "b": 2,
})


def a():
    print("a start")
    r = d.copy()
    for k, v in r.items():
        if k == "a":
            del r[k]
    print("a end")


def b():
    print("b start")
    r = d.copy()
    for k, v in r.items():
        if k == "b":
            del r[k]
    print("b end")


b()
a()
```
Output:
```
b start
b end
a start
  File "/home/tiger-222/projects/httpie/r.py", line 28, in <module>
    a()
  File test.py", line 12, in a
    for k, v in r.items():
  File "/.../Lib/_collections_abc.py", line 850, in __iter__
    for key in self._mapping:
  File "/.../requests/structures.py", line 60, in <genexpr>
    return (casedkey for casedkey, mappedvalue in self._store.values())
RuntimeError: OrderedDict mutated during iteration
```
We can see that we can alter the dict if the `del` happens last, and that's why we did not encounter the problem sooner. Maybe the situation is not so frequent.

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
   an alternative to ``stdin``. (`#534`_)
 * Fixed ``--continue --download`` with a single byte to be downloaded left. (`#1032`_)
 * Fixed ``--verbose`` HTTP 307 redirects with streamed request body. (`#1088`_)
+* Fixed session files handling when there are cookies and other headers following. (`#1126`_)
```
jakubroztocil, line 16:
```suggestion
* Fixed handling of session files with `Cookie:` followed by other headers. (`#1126`_)
```

tests/test_sessions.py
```diff
@@ -178,6 +178,25 @@ def test_session_by_path(self, httpbin):
         assert HTTP_OK in r2
         assert r2.json['headers']['Foo'] == 'Bar'
 
+    def test_session_with_cookie_followed_by_another_header(self, httpbin):
+        """Non-regression test for issue #1126 where request headers were mutated during iteration:
+
+            RuntimeError: OrderedDict mutated during iteration
+        """
```
jakubroztocil, line 185:
```suggestion
        """
        Make sure headers don’t get mutated — <https://github.com/httpie/httpie/issues/1126>
        
        """
```

jakubroztocil (changes requested):

tests/test_sessions.py
```diff
@@ -178,6 +178,25 @@ def test_session_by_path(self, httpbin):
         assert HTTP_OK in r2
         assert r2.json['headers']['Foo'] == 'Bar'
 
+    def test_session_with_cookie_followed_by_another_header(self, httpbin):
+        """
+        Make sure headers don’t get mutated — <https://github.com/httpie/httpie/issues/1126>
+
```
BoboTiG, line 184:
```suggestion
```

jakubroztocil:
🥧

PR #1130, opened 2021-08-18
Improve JSON output when there is leading data before the actual JSON body

BoboTiG:
In some special cases, to prevent against Cross Site Script Inclusion (XSSI) attacks, the JSON response body starts with a magic prefix line that must be stripped before feeding the rest of the response body to the JSON parser.
Such prefix is now simply ignored from the parser but still printed in the terminal.

Supersedes #529.

### DIFF
--- a/httpie/output/formatters/colors.py
+++ b/httpie/output/formatters/colors.py
@@ -9,10 +9,12 @@ import pygments.token
 from pygments.formatters.terminal import TerminalFormatter
 from pygments.formatters.terminal256 import Terminal256Formatter
 from pygments.lexer import Lexer
+from pygments.lexers.data import JsonLexer
 from pygments.lexers.special import TextLexer
 from pygments.lexers.text import HttpLexer as PygmentsHttpLexer
 from pygments.util import ClassNotFound
 
+from ..lexers.json import EnhancedJsonLexer
 from ...compat import is_windows
 from ...context import Environment
 from ...plugins import FormatterPlugin
@@ -60,6 +62,7 @@ class ColorFormatter(FormatterPlugin):
             http_lexer = PygmentsHttpLexer()
             formatter = TerminalFormatter()
         else:
+            from ..lexers.http import SimplifiedHTTPLexer
             http_lexer = SimplifiedHTTPLexer()
             formatter = Terminal256Formatter(
                 style=self.get_style_class(color_scheme)
@@ -151,55 +154,12 @@ def get_lexer(
         else:
             lexer = pygments.lexers.get_lexer_by_name('json')
 
+    # Use our own JSON lexer: it supports JSON bodies preceded by non-JSON data
+    # as well as legit JSON bodies.
+    if isinstance(lexer, JsonLexer):
+        lexer = EnhancedJsonLexer()
+
     return lexer
-
-
-class SimplifiedHTTPLexer(pygments.lexer.RegexLexer):
-    """Simplified HTTP lexer for Pygments.
-
-    It only operates on headers and provides a stronger contrast between
-    their names and values than the original one bundled with Pygments
-    (:class:`pygments.lexers.text import HttpLexer`), especially when
-    Solarized color scheme is used.
-
-    """
-    name = 'HTTP'
-    aliases = ['http']
-    filenames = ['*.http']
-    tokens = {
-        'root': [
-            # Request-Line
-            (r'([A-Z]+)( +)([^ ]+)( +)(HTTP)(/)(\d+\.\d+)',
-             pygments.lexer.bygroups(
-                 pygments.token.Name.Function,
-                 pygments.token.Text,
-                 pygments.token.Name.Namespace,
-                 pygments.token.Text,
-                 pygments.token.Keyword.Reserved,
-                 pygments.token.Operator,
-                 pygments.token.Number
-             )),
-            # Response Status-Line
-            (r'(HTTP)(/)(\d+\.\d+)( +)(\d{3})( +)(.+)',
-             pygments.lexer.bygroups(
-                 pygments.token.Keyword.Reserved,  # 'HTTP'
-                 pygments.token.Operator,  # '/'
-                 pygments.token.Number,  # Version
-                 pygments.token.Text,
-                 pygments.token.Number,  # Status code
-                 pygments.token.Text,
-                 pygments.token.Name.Exception,  # Reason
-             )),
-            # Header
-            (r'(.*?)( *)(:)( *)(.+)', pygments.lexer.bygroups(
-                pygments.token.Name.Attribute,  # Name
-                pygments.token.Text,
-                pygments.token.Operator,  # Colon
-                pygments.token.Text,
-                pygments.token.String  # Value
-            ))
-        ]
-    }
 
 
 class Solarized256Style(pygments.style.Style):
--- a/httpie/output/formatters/json.py
+++ b/httpie/output/formatters/json.py
@@ -17,15 +17,16 @@ class JSONFormatter(FormatterPlugin):
         ]
         if (self.kwargs['explicit_json']
                 or any(token in mime for token in maybe_json)):
+            from ..utils import load_prefixed_json
             try:
-                obj = json.loads(body)
+                data_prefix, json_obj = load_prefixed_json(body)
             except ValueError:
                 pass  # Invalid JSON, ignore.
             else:
                 # Indent, sort keys by name, and avoid
                 # unicode escapes to improve readability.
-                body = json.dumps(
-                    obj=obj,
+                body = data_prefix + json.dumps(
+                    obj=json_obj,
                     sort_keys=self.format_options['json']['sort_keys'],
                     ensure_ascii=False,
                     indent=self.format_options['json']['indent']
--- a/httpie/output/lexers/json.py
+++ b/httpie/output/lexers/json.py
@@ -1,0 +1,31 @@
+import re
+
+from pygments.lexer import bygroups, using, RegexLexer
+from pygments.lexers.data import JsonLexer
+from pygments.token import Token
+
+PREFIX_TOKEN = Token.Error
+PREFIX_REGEX = r'[^{\["]+'
+
+
+class EnhancedJsonLexer(RegexLexer):
+    """
+    Enhanced JSON lexer for Pygments.
+
+    It adds support for eventual data prefixing the actual JSON body.
+
+    """
+    name = 'JSON'
+    flags = re.IGNORECASE | re.DOTALL
+    tokens = {
+        'root': [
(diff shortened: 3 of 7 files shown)


httpie/output/formatters/json.py
```diff
@@ -1,14 +1,70 @@
 import json
+from typing import Generator
 
 from ...plugins import FormatterPlugin
 
+XXSI_PREFIXES = (")]}',\n", ")]}',", 'while(1);', 'for(;;)')
```
BoboTiG, line 6: I added the constant to ease future prefix add.

For example, I [read](http://patorjk.com/blog/2013/02/05/crafty-tricks-for-avoiding-xssi/) that some service was using `123\n` as prefix. Such prefix would break the formatting as no heuristics could handle it: it is only valid JSON values. For instance, that response's body would be problematic without the possibility to add `123\n` prefix later:
```json
123\n"some string"
```

For more context, have a look at the doctring of `JSONFormatter._find_potential_json_bodies()`.

httpie/output/formatters/json.py
```diff
@@ -1,14 +1,70 @@
...
+            yield data[min(token_position):]
+
+        # Finally, iterate over the data one character at a time.
+        # Note: that path should be rarely taken, it likely means
+        #       the data is a simple JSON string prefixed with a
+        #       custom XSSI prefix, or not JSON at all.
+        data = data[1:]
+        while data:
+            yield data
+            data = data[1:]
```
BoboTiG, line 54: This may be insane to do that. It is present only to handle simple JSON strings prefixed with custom XSSI string.

httpie/output/formatters/json.py
```diff
@@ -1,14 +1,57 @@
...
+            3. Finally, try to find the start of the actual body by applying some heuristics.
+
+        """
+        # First, the full data.
+        yield data
+
+        # Then, strip off eventual XSSI prefix.
+        for prefix in XXSI_PREFIXES:
+            if data.startswith(prefix):
+                yield data[len(prefix)]
```
BoboTiG, line 36:
```suggestion
                yield data[len(prefix):]
```

BoboTiG:
The prefix is printed before the prettified JSON:

Capture d’écran_2021-08-20_16-01-06

I am not sure yet how to set a specific color for the prefix @jakubroztocil. Do you have a clue :pray:?

jakubroztocil:
> I am not sure yet how to set a specific color for the prefix @jakubroztocil. Do you have a clue 🙏?

Take a look at what tokens Pygments works with. I’d probably experiment with comments, exceptions, errors, etc. Perhaps you could even generate an example for each token type to see what feels right?

https://github.com/objcio/pygments/blob/master/pygments/token.py#L115-L200

BoboTiG:
Here we are.

Style "error":
token-error

Style "keyword":
token-keyword

Style "punctuation":
token-punctuation

Style "string":
token-string

Actually, I prefer no color. But Maybe the "error" style is better to highlight data not being taken into account be the parser?

jakubroztocil:
From those four I’d pick error. But I’m wondering how the different comment tokens would look like. Could you write a loop over all the tokens and print the name and the output for each so that we can compare them?

tests/test_output.py
```diff
@@ -20,6 +21,8 @@
 from httpie.status import ExitStatus
 from .utils import COLOR, CRLF, HTTP_OK, MockEnvironment, http
 
+JSON_XXSI_PREFIXES = (")]}',\n", ")]}',", 'while(1);', 'for(;;)', ')', ']', '}')
```
jakubroztocil, line 24:
```suggestion
TEST_JSON_XXSI_PREFIXES = (")]}',\n", ")]}',", 'while(1);', 'for(;;)', ')', ']', '}')
```

(To make it clear these’re the ones we test, no an exhaustiveve list)

BoboTiG:
Here are all styles available (without duplicates):
tokens

BoboTiG:
Actually there are no "comment" token styles because JSON does not support comments.

BoboTiG:
I updated the PR, here is a short demo: [asciicast](https://asciinema.org/a/431804)

BoboTiG:
Both requests and responses are handled on purpose.

httpie/output/lexers/json.py
```diff
@@ -0,0 +1,44 @@
...
+                        yield index, self.body_prefix_token, char
+                    else:
+                        # Punctuation character outside the body prefix.
+                        yield index, token, char
+            elif in_body_prefix:
+                # `value` is part of the body prefix.
+                yield index, self.body_prefix_token, value
+            else:
+                # Outside the body prefix.
+                yield index, token, value
```
jakubroztocil, line 44: You can just have a simple lexer that delegates to the JSON one anything after the prefix (needs some testing). No additional treatment for the prefix is needed when it comes to colors, so you can simplify `ColorFormatter` again.

```python
from pygments.lexer import using, RegexLexer

class EnhancedJsonLexer(RegexLexer):
    flags = re.IGNORECASE | re.DOTALL
    tokens = {
        'root': [
            (r'[^{\[]+', Token.Error),  # Prefix
            (r'.+', using(JsonLexer)),
        ],
    }
```

httpie/output/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Sometimes the JSON body is prefixed with a XSSI magic string, specific to the server.
+    In order to be able to display well-formatted JSON, that method will try several
+    approaches to find the actual body of the JSON.
+
+    Yield tuples (data prefix, actual JSON body).
+
+    """
+    # First, the full data.
+    yield '', data
```
jakubroztocil, line 29: This should not be done here.  `load_prefixed_json()` should try to parse the whole string on its own before calling `parse_prefixed_json()`.

httpie/output/formatters/colors.py
```diff
@@ -78,13 +80,29 @@ def format_headers(self, headers: str) -> str:
...
+            for prefix, actual_body in find_json_actual_body_and_prefix(body):
+                continue
+
+            # `body` is prefixed with non-JSON data, likely a XSSI magic string.
+            # Use our own JSON lexer in order to properly display the data prefix
+            # and the actual JSON body.
+            if prefix:
+                from ..lexers.json import EnhancedJsonLexer
+                token = EnhancedJsonLexer.body_prefix_sep
+                body = f'{token}{prefix}{token}{actual_body}'
```
jakubroztocil, line 98: This is not needed. See my simplified JSON lexer below.

jakubroztocil (changes requested):

httpie/output/lexers/json.py
BoboTiG: Much simpler :+1:

jakubroztocil: My brain kept thinking about it in the background and the above probably breaks with naked strings, numbers, and booleans (we should have tests for those). A more explicit handling of the prefix should be more robust, something like:


```python
import re
from pygments.lexer import using, RegexLexer, bygroups


PrefixToken = Token.Error

class EnhancedJsonLexer(RegexLexer):
    flags = re.IGNORECASE | re.DOTALL
    tokens = {
        'root': [
            (r'([^{\[]+)([{\]].+)', bygroups(PrefixToken, using(JsonLexer))),
            (r'.+', using(JsonLexer)),
        ],
    }
```

BoboTiG:
I resolved the conflict, the patch is good for a final review :)

httpie/output/formatters/colors.py
```diff
@@ -78,13 +80,25 @@ def format_headers(self, headers: str) -> str:
...
+        # We may want to use a custom JSON lexer for special JSON data.
+        if type(lexer) is JsonLexer:
+            from ..utils import parse_prefixed_json
+            data_prefix = parse_prefixed_json(body)[0]
+            if data_prefix:
+                # `body` is prefixed with non-JSON data, likely a XSSI magic string.
+                # Use our own JSON lexer in order to properly display the data prefix
+                # and the actual JSON body.
+                from ..lexers.json import EnhancedJsonLexer
+                lexer = EnhancedJsonLexer()
```
jakubroztocil, line 95: You no longer need to check for the prefix. Just swap it. The custom lexer is meant to handle both scenarios.

jakubroztocil (changes requested):

httpie/output/formatters/colors.py
jakubroztocil: And the could be moved to our `get_lexer()`. https://github.com/httpie/httpie/blob/464b5b4c1d8b3e72255739094e17a013d80abae7/httpie/output/formatters/colors.py#L154

BoboTiG: And now that there is a way to override the `Content-Type`, do you agree to remove that block?
https://github.com/httpie/httpie/blob/464b5b4c1d8b3e72255739094e17a013d80abae7/httpie/output/formatters/colors.py#L145-L152

jakubroztocil: The idea behind the [explicit `--json` flag](https://httpie.io/docs#explicit-json) is to enable a maximally JSON-friendly mode. This extra detection is one of the features it enables (context: https://github.com/httpie/httpie/commit/74e4d0b67852ab0b628f687ac62e2d509fa74b61). 

With the new option for response content type overriding the user can achieve the same, but it takes more work and there’s a learning curve. 

The current behavior is not entirely elegant and I’m not sure how many people actually rely on it, but let’s leave it there at least for now, also for compatibility reasons.

tests/test_json.py
```diff
@@ -0,0 +1,39 @@
+import json
+
+import pytest
+import responses
+
+from httpie.cli.constants import PRETTY_MAP
```
BoboTiG, line 6:
```suggestion
from httpie.cli.constants import PRETTY_MAP
from httpie.compat import is_windows
```

tests/test_json.py
```diff
@@ -0,0 +1,39 @@
...
+from httpie.output.formatters.colors import ColorFormatter
+
+from .utils import MockEnvironment, http, URL_EXAMPLE
+
+TEST_JSON_XXSI_PREFIXES = (r")]}',\n", ")]}',", 'while(1);', 'for(;;)', ')', ']', '}')
+TEST_JSON_VALUES = ({}, {'a': 0, 'b': 0}, [], ['a', 'b'], 'foo', True, False, None)  # FIX: missing int & float
+TEST_PREFIX_TOKEN_COLOR = '\x1b[04m\x1b[91m'
```
BoboTiG, line 13:
```suggestion
TEST_PREFIX_TOKEN_COLOR = '\x1b[38;5;15m' if is_windows else '\x1b[04m\x1b[91m'
```

BoboTiG:
Merging the PR as it seems quite correct and all tests are green.

PR #1134, opened 2021-09-01
Allow to overwrite the response Content-Type from options

BoboTiG:
It implements the support for custom mime types.
It is a start, it will be used later in  #1110 to support custom a charset.

Summary:
```bash
$ http --response-as 'appllication/xml' \
    https://raw.githubusercontent.com/httpie/httpie/master/tests/fixtures/xmldata/valid/simple_raw.xml
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [2.6.0.dev0](https://github.com/httpie/httpie/compare/2.5.0...master) (unreleased)
 
 - Added support for formatting & coloring of JSON bodies preceded by non-JSON data (e.g., an XXSI prefix). ([#1130](https://github.com/httpie/httpie/issues/1130))
+- Added `--format-options response.content_type` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
+- Added `--response-content-type` shortcut for setting the response `Content-Type`-related `--format-options`. ([#1134](https://github.com/httpie/httpie/issues/1134))
 - Installed plugins are now listed in `--debug` output. ([#1165](https://github.com/httpie/httpie/issues/1165))
 - Fixed duplicate keys preservation of JSON data. ([#1163](https://github.com/httpie/httpie/issues/1163))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [2.6.0.dev0](https://github.com/httpie/httpie/compare/2.5.0...master) (unreleased)
 
 - Added support for formatting & coloring of JSON bodies preceded by non-JSON data (e.g., an XXSI prefix). ([#1130](https://github.com/httpie/httpie/issues/1130))
-- Added `--format-options response.as` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
+- Added `--format-options=response.as:CONTENT_TYPE` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
 - Added `--response-as` shortcut for setting the response `Content-Type`-related `--format-options`. ([#1134](https://github.com/httpie/httpie/issues/1134))
 - Installed plugins are now listed in `--debug` output. ([#1165](https://github.com/httpie/httpie/issues/1165))
 - Fixed duplicate keys preservation of JSON data. ([#1163](https://github.com/httpie/httpie/issues/1163))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [2.6.0.dev0](https://github.com/httpie/httpie/compare/2.5.0...master) (unreleased)
 
 - Added support for formatting & coloring of JSON bodies preceded by non-JSON data (e.g., an XXSI prefix). ([#1130](https://github.com/httpie/httpie/issues/1130))
-- Added `--format-options response.content_type` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
-- Added `--response-content-type` shortcut for setting the response `Content-Type`-related `--format-options`. ([#1134](https://github.com/httpie/httpie/issues/1134))
+- Added `--format-options response.as` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
+- Added `--response-as` shortcut for setting the response `Content-Type`-related `--format-options`. ([#1134](https://github.com/httpie/httpie/issues/1134))
 - Installed plugins are now listed in `--debug` output. ([#1165](https://github.com/httpie/httpie/issues/1165))
 - Fixed duplicate keys preservation of JSON data. ([#1163](https://github.com/httpie/httpie/issues/1163))
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1214,14 +1214,15 @@ You can further control the applied formatting via the more granular [format opt
 The `--format-options=opt1:value,opt2:value` option allows you to control how the output should be formatted
 when formatting is applied. The following options are available:
 
-|           Option | Default value | Shortcuts                |
-| ---------------: | :-----------: | ------------------------ |
-|   `headers.sort` |    `true`     | `--sorted`, `--unsorted` |
-|    `json.format` |    `true`     | N/A                      |
-|    `json.indent` |      `4`      | N/A                      |
-| `json.sort_keys` |    `true`     | `--sorted`, `--unsorted` |
-|     `xml.format` |    `true`     | N/A                      |
-|     `xml.indent` |      `2`      | N/A                      |
+|                  Option | Default value | Shortcuts                                           |
+| ----------------------: | :-----------: | --------------------------------------------------- |
+|          `headers.sort` |    `true`     | `--sorted`, `--unsorted`                            |
+|           `json.format` |    `true`     | N/A                                                 |
+|           `json.indent` |      `4`      | N/A                                                 |
+|        `json.sort_keys` |    `true`     | `--sorted`, `--unsorted`                            |
+| `response.content_type` |     `''`      | [`--response-content-type`](#response-content-type) |
+|            `xml.format` |    `true`     | N/A                                                 |
+|            `xml.indent` |      `2`      | N/A                                                 |
 
 For example, this is how you would disable the default header and JSON key
 sorting, and specify a custom JSON indent size:
@@ -1235,6 +1236,17 @@ sorting-related format options (currently it means JSON keys and headers):
 `--unsorted` and `--sorted`.
 
 This is something you will typically store as one of the default options in your [config](#config) file.
+
+#### Response `Content-Type`
+
+The `--response-content-type` option is a shortcut for `--format-options response.content_type`, and allows you to override the response `Content-Type`.
+It will be effective when the response is pretty-printed, and can be used when the server returns an incorrect `Content-Type` regarding the contents.
+
+For example, the following request will force the response to be treated as XML:
+
+```bash
+$ http --response-content-type 'application/xml' pie.dev/get
+```
 
 ### Binary data
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1239,13 +1239,14 @@ This is something you will typically store as one of the default options in your
 
 #### Response `Content-Type`
 
-The `--response-content-type` option is a shortcut for `--format-options response.content_type`, and allows you to override the response `Content-Type`.
-It will be effective when the response is pretty-printed, and can be used when the server returns an incorrect `Content-Type` regarding the contents.
+The `--response-content-type=value` option is a shortcut for `--format-options response.content_type:value`, 
+and it allows you to override the response `Content-Type` sent by the server. 
+That makes it possible for HTTPie to pretty-print the response even when the server specifies the type incorrectly.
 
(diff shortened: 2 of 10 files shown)


httpie/cli/constants.py
```diff
@@ -84,12 +84,13 @@
 }
 PRETTY_STDOUT_TTY_ONLY = object()
 
-
+FORMAT_OPTION_AUTOMATIC = 'automatic'
```
BoboTiG, line 87: It is generic enough to be used later by other format options (like I would do for #1110).

httpie/cli/definition.py
```diff
@@ -309,6 +309,22 @@
...
+    help='''
+        The response Content-Type can be specified explicitly via this option.
+        Useful when the Content-Type returned by the server is incorrect.
+        It is a shortcut for:
+
+            --format-options=response.content_type:CONTENT_TYPE
+
+        Example:
+
+            --response-content-type 'application/xml'
```
BoboTiG, line 324: In #1110 I will add more examples like:
```shell
--response-content-type 'application/xml; charset=utf-8'
--response-content-type 'charset=utf-8'
```

httpie/output/utils.py
```diff
@@ -0,0 +1,54 @@
+
+def parse_header(line):
+    """Parse a Content-Type like header.
+    Return the main Content-Type and a dictionary of options.
```
jakubroztocil, line 4: This looks like a specialized function with a very generic name. Something like `parse_structured_header()` or even `parse_content_type_header_overwrite()` would work better.

tests/test_xml.py
```diff
@@ -83,3 +83,57 @@ def test_invalid_xml(file):
...
+    # Request
+    r = http('--offline', '--raw', SAMPLE_XML_DATA,
+             '--format-options', 'response.content_type:application/xml',
+             URL_EXAMPLE)
+    assert SAMPLE_XML_DATA in r
+
+    # Response
+    r = http('--format-options', 'response.content_type:application/xml',
+             URL_EXAMPLE)
+    assert pretty_xml(parse_xml(SAMPLE_XML_DATA)) in r
```
jakubroztocil, line 106: Since this is used multiple times, let’s add a constant at the top, e.g.:

```python
XML_DATA_RAW = '<?xml version="1.0" encoding="utf-8"?><root><e>text</e></root>'
XML_DATA_FORMATTED = pretty_xml(parse_xml(SAMPLE_XML_DATA))
```

httpie/cli/constants.py
jakubroztocil: I think the default should simply be `null` (parsed as `None`).

jakubroztocil (changes requested):

README.rst
```diff
@@ -1580,6 +1580,8 @@ when formatting is applied. The following options are available:
 +--------------------+----------+---------------+------------------------------+
 | ``json.sort_keys``            | ``true``      | ``--sorted``, ``--unsorted`` |
 +--------------------+----------+---------------+------------------------------+
+| ``response.content_type``     | ``automatic`` | ``--response-content-type``  |
++--------------------+----------+---------------+------------------------------+
```
jakubroztocil, line 1584: This feature will need an explanation in the docs. I’m also not sure about the naming — what this allows you to do is to "overwrite the response’s `Content-Type` for output formatting purposes". It’s a tricky one to capture under a short name, but let’s think about this a bit more.

httpie/output/streams.py
```diff
@@ -136,7 +138,16 @@ def __init__(
         super().__init__(**kwargs)
         self.formatting = formatting
         self.conversion = conversion
-        self.mime = self.msg.content_type.split(';')[0]
+        self.msg_type = 'request' if isinstance(self.msg, HTTPRequest) else 'response'
```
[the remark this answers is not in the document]
BoboTiG: For now. Today we introduce `response` content type. Tomorrow we may want to add one for `request`. As stated in the description, that attribute will be useful for #1110 too.

Another way would be to introduce specialized classes, like `PrettyStreamRequest` and `PrettyStreamResponse`, but that would be too much for such thing and would require changing more code "only" to make the distinction about responses and requests.

httpie/cli/constants.py
BoboTiG: Sure.

httpie/output/processing.py
```diff
@@ -51,3 +52,30 @@ def format_body(self, content: str, mime: str) -> str:
             for p in self.enabled_plugins:
                 content = p.format_body(content, mime)
         return content
+
+    def option(self, *names: List[str], default: Any = None) -> Any:
```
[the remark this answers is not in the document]
BoboTiG: It think it is more convenient to do:
```python
self.option('request, 'content_type', default=FORMAT_OPTION_AUTOMATIC)
```
than:
```python
self.format_options.get('request', {}).get('content_type', FORMAT_OPTION_AUTOMATIC)
```
`default` is "needed" now that we introduce `request` and `response` notions. It is interesting to have it now and for the future.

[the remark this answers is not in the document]
jakubroztocil: All the options are predefined. They’re guaranteed to exist and each of them has a default value. 

You don’t need to access it via:

```python
self.format_options.get('request', {}).get('content_type', FORMAT_OPTION_AUTOMATIC)
```

But simply as:

```python
self.format_options['request']['content_type']
````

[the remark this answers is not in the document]
Comment (BoboTiG) [reply httpie/output/processing.py:56]: With other changes requested, `default` me become obsolete though. We could simply return `None` when an option is not available.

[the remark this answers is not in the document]
Comment (BoboTiG) [reply httpie/output/processing.py:56]: (I did not refresh the page so I did not see your last response).

My point is more:
```python
self.format_options[self.msg_type]['content_type']
```
where `self.msg_type` is not an option.

README.rst
BoboTiG: And it is already too long IMO. I am not sure adding more keywords would be better :thinking:

BoboTiG: I think "formatting purposes" and "overwrite" should not be part of such option."
Same as `--sorted` does not reflect that, else we would end up with `--force-sorting-response-and-headers-json-keys`. I made it to make my point, but indeed such specialized option could be renamed. Let's think about it.

httpie/output/streams.py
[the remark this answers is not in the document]
jakubroztocil: It’s not clear that the content type override makes sense for requests. I’d prefer not complicating the code preemptively. My main issue with the code is that hard-coded strings with no meaningful names attached to them are fragile. If we needed this, then adding something `HTTPMessage.type_name` would be more robust and you could easily find all the places where we touch that.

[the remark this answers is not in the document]
BoboTiG: That is another option, yes :+1:

README.rst
jakubroztocil: Yeah, I would definitely not make it longer.

httpie/output/processing.py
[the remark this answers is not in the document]
BoboTiG: I will remove all that code and do a simple `if` to apply the formatting only for responses. Let's rework that part when we will really need to handle more specialized cases.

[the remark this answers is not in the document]
jakubroztocil: Sounds good. I think there’s a lot of value in keeping it a simple `dict` rather then inventing a custom interface for the lookups. We could even type-annotate the `dict` that holds formatting options.

httpie/cli/definition.py
```diff
@@ -309,6 +309,21 @@
     '''
 )
 
+output_processing.add_argument(
+    '--response-content-type',
+    metavar='CONTENT_TYPE',
+    help='''
+        Overwrite the response Content-Type. Useful when the Content-Type returned
+        by the server is incorrect. It is a shortcut for:
```
BoboTiG, line 317: Maybe "Useful when the Content-Type returned by the server is incorrect." should be removed, WDYT?

httpie/output/streams.py
```diff
@@ -136,7 +137,15 @@ def __init__(
         super().__init__(**kwargs)
         self.formatting = formatting
         self.conversion = conversion
-        self.mime = self.msg.content_type.split(';')[0]
+        self.mime = self.get_mime()
+
+    def get_mime(self) -> str:
+        mime = parse_header_content_type(self.msg.content_type)[0]
+        if self.msg.type_name == 'response':
```
jakubroztocil, line 144: I’d much prefer the instance check here. And if this is the only place where we touch `msg.type_name`, then we don’t really need the attribute.

httpie/cli/definition.py
jakubroztocil: I think mentioning the main use case is a good idea. I’d also explain that it’s for display/formatting purposes only. But let’s look at the overall documentation together, I have some more ideas and we need to coordinate the `--help` output with README.

httpie/cli/argtypes.py
```diff
@@ -225,7 +226,11 @@ def parse_format_options(s: str, defaults: Optional[dict]) -> dict:
                     f'invalid key {path!r}')
 
             default_type, parsed_type = type(default_value), type(parsed_value)
-            if parsed_type is not default_type:
+            if default_value is None:
+                # It is allowed to change an option when its default
+                # value is 'null' (parsed as `None`).
+                pass
```
jakubroztocil, line 232: This makes me think that maybe using `''` as the default value could be cleaner. What do you think?

BoboTiG: :thinking: 

Using nothing:
```
  --format-options FORMAT_OPTIONS
      (...)
      
          headers.sort:true
          json.format:true
          json.indent:4
          json.sort_keys:true
          response.content_type:
          xml.format:true
          xml.indent:2
```

Using `''`:
```
  --format-options FORMAT_OPTIONS
      (...)
      
          headers.sort:true
          json.format:true
          json.indent:4
          json.sort_keys:true
          response.content_type:''
          xml.format:true
          xml.indent:2
```

Using `null`:
```
  --format-options FORMAT_OPTIONS
      (...)
      
          headers.sort:true
          json.format:true
          json.indent:4
          json.sort_keys:true
          response.content_type:null
          xml.format:true
          xml.indent:2
```

`''` has the advantage of the type check being the same (always `str`). And actually it renders better,  I think. `null` seems like a bad value not well translated.

Let's go with `''` :+1:

docs/README.md
```diff
@@ -1236,6 +1237,17 @@ sorting-related format options (currently it means JSON keys and headers):
...
 
+#### Response `Content-Type`
+
+The `--response-content-type` option is a shortcut for `--format-options response.content_type`, and allows you to override the response `Content-Type`.
+It will be effective when the response is pretty-printed, and can be used when the server returns an incorrect `Content-Type` regarding the contents.
+
+For example, the following request will force the response to be treated as XML:
+
+```bash
+$ http --response-content-type 'application/xml' pie.dev/get
```
jakubroztocil, line 1248:
```suggestion
$ http --response-content-type=application/xml pie.dev/get
```

docs/README.md
```diff
@@ -1236,6 +1237,17 @@ sorting-related format options (currently it means JSON keys and headers):
 
 This is something you will typically store as one of the default options in your [config](#config) file.
 
+#### Response `Content-Type`
+
+The `--response-content-type` option is a shortcut for `--format-options response.content_type`, and allows you to override the response `Content-Type`.
+It will be effective when the response is pretty-printed, and can be used when the server returns an incorrect `Content-Type` regarding the contents.
```
jakubroztocil, line 1243:
```suggestion
The `--response-content-type=value` option is a shortcut for `--format-options response.content_type:value`, 
and it allows you to override the response `Content-Type` sent by the server. 
That makes it possible for HTTPie to pretty-print the response even when the server specifies the type incorrectly.
```

httpie/cli/definition.py
```diff
@@ -309,6 +309,18 @@
     '''
 )
 
+output_processing.add_argument(
+    '--response-content-type',
+    metavar='CONTENT_TYPE',
+    help='''
+    Override the response Content-Type. It will be effective when the response
+    is pretty-printed, and can be used when the server returns an incorrect
+    Content-Type regarding the contents. It is a shortcut for:
```
jakubroztocil, line 318:
```suggestion
    Override the response Content-Type for formatting purposes, e.g.:
    
        --response-content-type=application/xml
    
    It is a shortcut for:
```

jakubroztocil (commented):
How about `--response-as` / `--treat-response-as`?

BoboTiG:
What would give both options:
```
--response-content-type='application/xml; charset=utf-8'
```
```
--response-as='application/xml; charset=utf-8'
```

:thinking:

I kind of like the `--response-as`. We could go with it if it is OK for you too when there is the charset specified also.

jakubroztocil:
@BoboTiG I think `--response-as=<content-type>` works for the `charset` use case as well. Let’s go with it: if we end up not liking it, we can still easily change the name before the next release. Btw, you’ll be adding the charset docs for this option once the #1110 is finished, right?

BoboTiG:
Yes, charset stuff has to be done with #1110.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [2.6.0.dev0](https://github.com/httpie/httpie/compare/2.5.0...master) (unreleased)
 
 - Added support for formatting & coloring of JSON bodies preceded by non-JSON data (e.g., an XXSI prefix). ([#1130](https://github.com/httpie/httpie/issues/1130))
+- Added `--format-options response.as` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
```
BoboTiG, line 9:
```suggestion
- Added `--format-options=response.as:CONTENT_TYPE` to allow overriding the response `Content-Type`. ([#1134](https://github.com/httpie/httpie/issues/1134))
```

PR #1136, opened 2021-09-02
Fix support for formatting of JSON with duplicate keys

### LINKED ISSUE #1133: Don't silently fix invalid JSON with duplicate keys
**Checklist**

- [x] I've searched for similar feature requests.

---

**What enhancement would you like to see?**

When a server returns invalid JSON, the HTTPie formatter should either leave it untouched or fail loudly. As it stands, if the JSON contains duplicate keys, HTTPie will quietly remove the duplicates during formatting.

**What problem does it solve?**

When debugging broken services, it's important that I can trust HTTPie to faithfully show me what the server returned, so that I focus on figuring out what's going wrong with my services and not have to spend that time thinking about HTTPie.

From my point of view it's helpful for HTTPie to format and color text, I'd rather it didn't sort keys by default but I know that I can reconfigure that, but it's a problem if HTTPie is removing anything that isn't non-significant whitespace. If HTTPie is to be a general purpose tool it is important that it doesn't mask problems.

**Provide any additional information, screenshots, or code examples below:**

The below example is my motivation for this feature. This server is returning invalid JSON with duplicate keys in the "dps" object. But I could not tell it was doing this, because HTTPie had silently "corrected" the output during formatting. You can see that if I turn off formatting and only colorize the output, I see the true output.

```
annettewilson@ANNEWILS02M:~/scratch/undercity$ http GET 'http://opentsdb.skymetrics.prod.skyscanner.local:4242/api/query' m=='zimsum:prod.undercity.search_cycle.meter.count_delta{}{dataCenter=literal_or(EU_WEST_1),endpoint=literal_or(search),from_bot=literal_or(false)}' start==1630064720 end==1630064727
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 232
Content-Type: application/json;charset=utf-8
Date: Wed, 01 Sep 2021 09:07:45 GMT
Vary: Accept-Encoding
X-OpenTSDB-Query-Complexity: 318

[
    {
        "aggregateTags": [
            "ip",
            "traffic_source"
        ],
        "dps": {
            "1630064726": 6.0
        },
        "metric": "prod.undercity.search_cycle.meter.count_delta",
        "tags": {
            "dataCenter": "EU_WEST_1",
            "endpoint": "search",
            "from_bot": "false"
        }
    }
]


annettewilson@ANNEWILS02M:~/scratch/undercity$ http GET 'http://opentsdb.skymetrics.prod.skyscanner.local:4242/api/query' m=='zimsum:prod.undercity.search_cycle.meter.count_delta{}{dataCenter=literal_or(EU_WEST_1),endpoint=literal_or(search),from_bot=literal_or(false)}' start==1630064720 end==1630064727 --pretty colors
HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
Date: Wed, 01 Sep 2021 09:07:59 GMT
Vary: Accept-Encoding
X-OpenTSDB-Query-Complexity: 318
Content-Length: 232
Connection: keep-alive

[{"metric":"prod.undercity.search_cycle.meter.count_delta","tags":{"endpoint":"search","dataCenter":"EU_WEST_1","from_bot":"false"},"aggregateTags":["ip","traffic_source"],"dps":{"1630064726":5.0,"1630064726":3.0,"1630064726":6.0}}]
```

BoboTiG:
Fixes #1133.


httpie/utils.py
```diff
@@ -12,10 +12,29 @@
...
+
+    def items(self):
+        return self._items
+
+
 def load_json_preserve_order(s):
     return json.loads(s, object_pairs_hook=OrderedDict)
 
 
+def load_json_preserve_duplicate_keys(s):
```
BoboTiG, line 34: I hesitated to alter `load_json_preserve_order()` directly. LMK if you think it is better to do it.

BoboTiG: IMO it is good as-is. Let's not alter too much the behavior.

jakubroztocil: Looking at where we use `load_json_preserve_order()` (with `=@` and `:=`), I think it makes sense to preserve the duplicate keys there as well. We never interpret the JSON body data; we only either send/save it somewhere or print it on the terminal. So we shouldn’t apply any cleanup (like the default duplicate key removal).

BoboTiG:
Hmm Python 3.6 and 3.7 failures are legit and will be tough to tackle. The problem is how `sorted()` is used within those versions.

- Python 3.6 and 3.7: https://github.com/python/cpython/blob/79101b890ee021a901a8b6837a3a320d57adb725/Lib/json/encoder.py#L353
- Python 3.8+: https://github.com/python/cpython/blob/9d318ad3ecca69b33fb4bb23a308488ec68e82f0/Lib/json/encoder.py#L353

I tracked the change and commented on https://bugs.python.org/issue23493, let's see what will happen next.

In the meantime, handling all Python versions equally is not possible without rewriting the JSON encoder. Either we add support for duplicate keys for Python 3.8+, either we loose JSON keys sorting for Python 3.6 & 3.7. The later is a breaking change, we do not want that :fire: 

I do not like having different behavior given the Python version, but 3.6 & 3.7 are already old and I would be OK to introduce duplicate keys support starting with 3.8.

WDYT @jakubroztocil ?

httpie/utils.py
```diff
@@ -12,10 +12,29 @@
...
 
+class JsonDictWithDuplicateKeys(OrderedDict):
+    """A specialized JSON dict with duplicate keys support.
+
+    """
+
+    def __init__(self, items):
+        # `json.dumps()` will check the underlying dict (`self` in our case) and outputs
+        # directly if it is empty. So we need to have something in the dictionary.
+        self[''] = ''
```
jakubroztocil, line 23: Is there a bool check somewhere? I think something like this with a docstring would be cleaner — as far as a hack of this type goes:

```python
def __bool__(self):
  return True
```

tests/fixtures/test_with_dupe_keys.json
```diff
@@ -0,0 +1,7 @@
+{
+    "dps": {
+        "1630064726": 5.0,
+        "1630064726": 3.0,
+        "1630064726": 6.0
```
jakubroztocil, line 5: Could we use simpler keys/values? To make it easier to scan.

httpie/utils.py
```diff
@@ -12,10 +12,29 @@
 RE_COOKIE_SPLIT = re.compile(r', (?=[^ ;]+=)')
 
 
+class JsonDictWithDuplicateKeys(OrderedDict):
```
jakubroztocil, line 15: Maybe instead of `with` something like `allowing` or `supporting`.

httpie/utils.py
BoboTiG: That seems to work, thanks! :champagne:

httpie/utils.py
jakubroztocil: …or actually`preserving` might the most fitting :)

BoboTiG: So it will be ;)

httpie/utils.py
BoboTiG: Finally, that is not enough. Overriding `__bool__` works only for the Python implementation of the JSON encoder. It will do a simple [`if not dct` check](https://github.com/python/cpython/blob/9d318ad3ecca69b33fb4bb23a308488ec68e82f0/Lib/json/encoder.py#L334-L336).

But most of the time, it will be the C implementation of the JSON encoder that will be used. And the check is not quite the same: it does [` if (PyDict_GET_SIZE(dct) == 0)`](https://github.com/python/cpython/blob/9d318ad3ecca69b33fb4bb23a308488ec68e82f0/Modules/_json.c#L1581-L1582), which is [defined like](https://github.com/python/cpython/blob/f8a95df84bcedebc0aa7132b3d1a4e8f000914bc/Include/cpython/dictobject.h#L48):
```c
#define PyDict_GET_SIZE(mp)  (assert(PyDict_Check(mp)),((PyDictObject *)mp)->ma_used)
```
The problem here, is that `ma_used` is updated only when an item is effectively added to the dict, and I am not aware of a workaround to fake that value.

BoboTiG: Is it worth adding so much technical details to the comment below `self[''] = ''`? I would say yes, WDYT?

jakubroztocil: Right. Is there’s anything special about using specifically `''`? Maybe we could put there something more explicit like `self.['__hack__'] = '__hack__'` (also easier to debug if it ever appears in some output).

BoboTiG: No, I just used the smaller string available. I'll go with `__hack__` :+1:

httpie/utils.py
```diff
@@ -12,8 +12,46 @@
 RE_COOKIE_SPLIT = re.compile(r', (?=[^ ;]+=)')
 
 
-def load_json_preserve_order(s):
-    return json.loads(s, object_pairs_hook=OrderedDict)
+class JsonDictPreservingDuplicateKeys(OrderedDict):
```
BoboTiG, line 15: I will use `dict` parent class and add a custom `__repr__()`. It should be cleaner.

PR #1143, opened 2021-09-06
Add GFM plugin for mdformat

hukkin:
I noticed that you are using mdformat and have GFM tables in the docs. The GFM plugin is needed in this case to parse (and format) GFM specific syntax properly.


jakubroztocil:
Thanks @hukkin, we’ll take a look. (For posterity: GFM — [GitHub Flavored Markdown](https://github.github.com/gfm/).)

hukkin:
:+1: 

BTW, if you ever decide to format GitHub issue templates too, you'll find out that you'll need [mdformat-frontmatter](https://pypi.org/project/mdformat_frontmatter/) also, as the frontmatter syntax used in those files is part of neither CommonMark nor GFM specification.

BoboTiG:
Thanks @hukkin, but we moved away from `mdformat`. Therefore I am closing the issue :)

PR #1145, opened 2021-09-07
Packit: Get the current Fedora Rawhide specfile

hroncok:
Using the fork is not needed anymore,
since Rawhide was updated to 2.5.0 and no longer has patches.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,10 +2,8 @@
 # https://packit.dev/docs/configuration/
 specfile_path: httpie.spec
 actions:
-  # the current Fedora Rawhide specfile has some patches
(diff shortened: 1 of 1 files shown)


BoboTiG (approved):

hroncok:
The rawhide build failure is transient.

BoboTiG:
Oh OK, so let'se merge it now :+1:

hroncok:
Does that happen often? If so, we could revert https://github.com/httpie/httpie/pull/1119

BoboTiG:
Thank you @hroncok :pie:

BoboTiG:
No, it is quite reliable actually.

PR #1146, opened 2021-09-07
use LZO compression for snap

om26er:
The snap archive should be compressed using LZO compression. This makes things relatively fast. 

Some reading on why https://snapcraft.io/blog/why-lzo-was-chosen-as-the-new-compression-method


BoboTiG (approved):

BoboTiG:
Great, thanks @om26er! :pie:

PR #1149, opened 2021-09-08
Split the monolithic workflow into specific ones

BoboTiG:


jakubroztocil:
Good idea 💡 Minor: I’d vote for using `-` instead of `_` as a word separator in all the action file names.

BoboTiG:
It works for me :+1:

PR #1150, opened 2021-09-08
Use HTTPie for the documentation build request

BoboTiG:
~WIP until HTTPie will be available on snap.~

### DIFF
--- a/.github/workflows/update-documentation.yml
+++ b/.github/workflows/update-documentation.yml
@@ -16,5 +16,7 @@ jobs:
   trigger-doc-build:
     runs-on: ubuntu-latest
     steps:
+      - name: Install HTTPie
+        run: sudo snap install --edge httpie
       - name: Trigger new documentation build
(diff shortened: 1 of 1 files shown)


.github/workflows/update-documentation.yml
```diff
@@ -16,5 +16,7 @@ jobs:
   trigger-doc-build:
     runs-on: ubuntu-latest
     steps:
+      - name: Install HTTPie
+        run: sudo snap install --edge httpie
       - name: Trigger new documentation build
-        run: curl -X POST ${{ secrets.DOCS_UPDATE_VERCEL_HOOK }}
+        run: http POST ${{ secrets.DOCS_UPDATE_VERCEL_HOOK }}
```
jakubroztocil, line 22: I think we need `--ignore-stdin` — https://httpie.io/docs#scripting:

```suggestion
        run: http --ignore-stdin POST ${{ secrets.DOCS_UPDATE_VERCEL_HOOK }}
```

PR #1154, opened 2021-09-10
[snap] Remove personal-files interface

BoboTiG:
Use of the `personal-files` interface is reserved for vetted publishers.

The interface requires a validation, but we need to publish at least one package first. So let's skip that part, release a version and ask for the interface access in a second time.

Also add a workflow to build & test the snap package.

### DIFF
--- a/.github/workflows/packaging-linux-snap.yml
+++ b/.github/workflows/packaging-linux-snap.yml
@@ -1,0 +1,25 @@
+name: Linux snap
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/packaging-linux-snap.yml
+      - snapcraft.yaml
+
+jobs:
+  snap:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build
+        uses: snapcore/action-build@v1
+        id: snapcraft
+      - name: Install
+        run: sudo snap install --dangerous ${{ steps.snapcraft.outputs.snap }}
+      - name: Test
+        run: |
+          httpie.http --version
+          httpie.https --version
+          # Auto-aliases cannot be tested when installing a snap outside the store.
+          # http --version
+          # https --version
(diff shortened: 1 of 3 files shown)


jakubroztocil:
HTTPie is not going to work at all without access to the config file. The sessions feature won’t work either. 

```
~ $ chmod 000 ~/.config/httpie
~ $ http
  File "/opt/homebrew/bin/http", line 8, in <module>
    sys.exit(main())
  File "/opt/homebrew/lib/python3.9/site-packages/httpie/__main__.py", line 11, in main
    exit_status = main()
  File "/opt/homebrew/lib/python3.9/site-packages/httpie/core.py", line 39, in main
    if env.config.default_options:
  File "/opt/homebrew/lib/python3.9/site-packages/httpie/context.py", line 109, in config
    if not config.is_new():
  File "/opt/homebrew/lib/python3.9/site-packages/httpie/config.py", line 82, in is_new
    return not self.path.exists()
  File "/opt/homebrew/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1414, in exists
    self.stat()
  File "/opt/homebrew/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1222, in stat
    return self._accessor.stat(self)
PermissionError: [Errno 13] Permission denied: '/Users/jakub/.config/httpie/config.json'
```

BoboTiG:
Ah a second too late!

This will still work because Snap is mocking `$HOME`. Config and sessions files will not work at all, but this is the case for all snap packages released so far. So snap users will not be in troubles.

BoboTiG:
I want to fix it after a first release to be able to ask for the interface access. But even with that access, we will only have a small amount of the feature: for instance `HTTPIE_CONFIG_DIR` is never gonna work if it points to `/tmp` or `~/$HOME`.

PR #1157, opened 2021-09-15
[snap] Improve OS integration

BoboTiG:
Get back read-write access to `$HOME/.httpie` and `$HOME/.config/httpie`.


BoboTiG:
Snap servers are under heavy load, that explains the job failure. Will wait before merging.

PR #1163, opened 2021-09-21
Fix duplicate keys preservation of JSON data

### LINKED ISSUE #1133: Don't silently fix invalid JSON with duplicate keys

BoboTiG:
Supersedes #1136 (I messed up with remote repositories).

Fixes #1133.


httpie/utils.py
```diff
@@ -1,19 +1,72 @@
...
+Items = List[Item]
+
 
+class JsonDictPreservingDuplicateKeys(OrderedDict):
+    """A specialized JSON dict preserving duplicate keys.
+
+    Note: Python 3.6 & 3.7 do not support duplicate keys sorting.
+          See https://bugs.python.org/issue23493#msg400929.
+
+    """
```
jakubroztocil, line 24: What do you mean exactly by "duplicate key sorting"?

jakubroztocil: I see that you check the Python version in the tests and refer from there to this docstring. It would be better to add a constant and explain everything in a comment next to it instead of the version checks and repeated explanations.

```python
JsonDictPreservingDuplicateKeys:
    # …
    SUPPORTS_SORTING = sys.version_info >= (3, 8)
```

httpie/utils.py
```diff
@@ -1,19 +1,72 @@
...
+        # is updated only when an item is added/removed from the dict. For that case,
+        # there is no workaround but to add an item into the dict.
+        # Sources:
+        #     - https://github.com/python/cpython/blob/9d318ad/Modules/_json.c#L1581-L1582
+        #     - https://github.com/python/cpython/blob/9d318ad/Include/cpython/dictobject.h#L53
+        #     - https://github.com/python/cpython/blob/9d318ad/Include/cpython/dictobject.h#L17-L18
+        #
+        # To please both implementations, we simply add one item to the dict.
+        if items:
+            self['__hack__'] = '__hack__'
```
jakubroztocil, line 49: I’d move the hack to a function to isolate it and have room for a docstring and a cleaner constructor.

```suggestion
    def __init__(self, items: Items):
        self._items = items
        self._ensure_items_used()

    def _ensure_items_used(self):
        """"
        HACK: Force `json.dumps()` to use `self.items()` instead of an empty dict.
        
        Two JSON encoders are available on CPython: pure-Python (1) and C (2) implementations.
        
        (1) The pure-python implementation will do a simple `if not dict: return '{}'`,
        and we could fake that check by implementing the `__bool__()` method.
        Source:
            - https://github.com/python/cpython/blob/9d318ad/Lib/json/encoder.py#L334-L336
        
        (2) On the other hand, the C implementation will do a check on the number of
        items contained inside the dict, using a verification on `dict->ma_used`, which
        is updated only when an item is added/removed from the dict. For that case,
        there is no workaround but to add an item into the dict.
        Sources:
            - https://github.com/python/cpython/blob/9d318ad/Modules/_json.c#L1581-L1582
            - https://github.com/python/cpython/blob/9d318ad/Include/cpython/dictobject.h#L53
            - https://github.com/python/cpython/blob/9d318ad/Include/cpython/dictobject.h#L17-L18
        #
        To please both implementations, we simply add one item to the dict.
        """
        if items:
            self['__hack__'] = '__hack__'
```

jakubroztocil (changes requested):

httpie/utils.py
jakubroztocil: A minor thing: let’s wrap all URLs in comments in angled braces like this:

```
See <https://bugs.python.org/issue23493#msg400929>.
```

That makes URLs visually stand out, easier to copy, and harder to accidentally change, especially when used in a sentence like above.

httpie/utils.py
```diff
@@ -1,19 +1,72 @@
...
+        """Return all items, duplicate ones included.
+
+        """
+        return self._items
+
+    def __eq__(self, other: object) -> bool:
+        """Most simple way to check for equality without reimplementing a complex logic.
+
+        """
+        return repr(self) == repr(other)
```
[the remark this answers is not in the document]
BoboTiG: It is to please our tests that compare a `dict` against a `JsonDictPreservingDuplicateKeys`.
I will remove the method and fix tests instead.

PR #1169, opened 2021-10-01
Add support for nested JSON syntax

BoboTiG:
## Proposal

Data fields (`key=value`) and raw JSON fields (`key:=value`) now parse the `key` as [JSON form](https://www.w3.org/TR/html-json-forms/).

Example:
```bash
$ http PUT pie.dev/put \
    'object=scalar' \                           # Object — blank key
    'object[0]=array 1' \                       # Object — "0" key
    'object[key]=key key' \                     # Object — "key" key
    'array:=1' \                                # Array — first item
    'array:=2' \                                # Array — second item
    'array[]:=3' \                              # Array — append (third item)
    'wow[such][deep][3][much][power][!]=Amaze'  # Nested object
```
```json
{
    "array": [
        1,
        2,
        3
    ],
    "object": {
        "": "scalar",
        "0": "array 1",
        "key": "key key"
    },
    "wow": {
        "such": {
            "deep": [
                null,
                null,
                null,
                {
                    "much": {
                        "power": {
                            "!": "Amaze"
                        }
                    }
                }
            ]
        }
    }
}
```

The new syntax is compatible with other JSON-specific syntax and can be mixed altogether.

---

Note: it is not doable to simply use the `jarg` module as it would install PyYAML, and we want to keep our requirements light. I also checked other modules like [html-json-forms](https://github.com/wq/html-json-forms) but I prefer to ship our own code for the same reason: no new module to package. At the end, the new code is maintainable enough, and *quite* small.


isidentical:
Merging this to `feature/nested-json`.

PR #1175, opened 2021-10-07
Added the ability to silence warnings via double `-q` or `--quiet`

dkreeft:


BoboTiG:
Thanks @dkreeft, that's great work :) :pie:

PR #1178, opened 2021-10-09
change elif to the start function

nixbytes:
refactor the filename_from_url with elif to simplify the code a little. ran the test and make came out clean. hope this helps.


jakubroztocil:
Thanks, @nixbytes! This would generally be a  good cleanup, but those conditions are done in that slightly more verbose way intentionally — to make the flow more explicit/easier to follow.

nixbytes:
oh ok, I understand. I will look around and see if there is any other section I could help with

PR #1183, opened 2021-10-15
Do not append missing XML declaration in formatted responses

### LINKED ISSUE #1156: Do not append missing XML declaration in formatted responses

1. Create the file `repro.xml` with that content:

  ```xml
  <!-- comment --><root><element key='value'>text</element><element>text</element>tail<empty-element/></root>
  ```
2. Serve that file using:

  ```bash
  $ python -m http.server
  ```
3. And GET the file:

  ```bash
  $ http :8000/repro.xml
  ```

## Current result

```xml
<?xml version="1.0" encoding="utf-8"?>
<!-- comment -->
<root>
  <element key="value">text</element>
  <element>text</element>
  tail
  <empty-element/>
</root>
```

Note the additional XML declaration (it is not present from the original response).

## Expected result

The responses should not be altered:

```xml
<!-- comment -->
<root>
  <element key="value">text</element>
  <element>text</element>
  tail
  <empty-element/>
</root>
```

---

When pretty-printing a XML response that does not contain a XML declaration, `minidom` will append one ([source](https://github.com/python/cpython/blob/851811f5772c43f72f445e2ce1ac3ea9da951ae3/Lib/xml/dom/minidom.py#L1823)).

this-is-r-gaurav:
Fixes #1156.

### DIFF
--- a/httpie/output/formatters/xml.py
+++ b/httpie/output/formatters/xml.py
@@ -43,22 +43,18 @@ class XMLFormatter(FormatterPlugin):
 
         from xml.parsers.expat import ExpatError
         from defusedxml.common import DefusedXmlException
-        pretty_body = body
         try:
             parsed_body = parse_xml(body)
         except ExpatError:
             pass  # Invalid XML, ignore.
         except DefusedXmlException:
             pass  # Unsafe XML, ignore.
-        else:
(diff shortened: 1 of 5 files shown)


isidentical:
@this-is-r-gaurav thanks for the PR, but we decided to with another way (#1227).

PR #1184, opened 2021-10-15
dnf/yum update is the same as dnf upgrade -- it updates all packages

hroncok:
No reason to run it before installing or upgrading httpie.
This is not apt.


hroncok:
Added here together with the READMe changes, so it comes in a single commit.

BoboTiG:
Perfect!

Thanks a lot :) :pie:

PR #1200, opened 2021-11-07
cmd: Implement httpie plugins interface

### LINKED ISSUE #566: Brew installed version doesn't work with plugins

isidentical:
Resolves #566.

Remaining stuff:
- [x] Fix windows (apparently some sysconfig issues)
- [x] Docs
- [x] More tests (and possibly some functional tests for error scenerios)
- [x] Playing with real world plugins (not part of the PR, but we should test them and ensure they are working both ways)
- [x] Improving UI for errors (e.g PermissionError)


isidentical:
The testing bit took a lot more than I thought, but I think it worths it (since it simplifies workflow by a lot). Now focusing on docs / remaining smaller tests.

isidentical:
I'll take a look at the package builds at last, but I feel like it should be easy to resolve.

isidentical:
Here is the UI demo for the current version;
[asciicast](https://asciinema.org/a/yoEQSekZMKjrtlinyFf1cCsvy)

isidentical:
I think I need to send a pull request to https://src.fedoraproject.org/rpms/httpie/blob/rawhide/f/httpie.spec after this gets merged since packit uses that file for the package configuration.

jakubroztocil (changes requested):
I’m yet to dive to the nitty-gritty, but overall it looks pretty solid 👍🏻 

One thing we should definitely improve though is the UX of when the management `httpie` command gets invoked without args or with invalid args:

The thing is that _many_ people (my guess would be ~50%+, i.e., higher thousands a month) will, after running `x install httpie`, run `httpie`. 

We need to _anticipate_ that and give them a crystal clear indication about the command/project name mismatch and help them get towards sending their first request; the mismatch is highly unintuitive.  (The docs could also be more clear in this sense.)  And the ability to direct people to the correct command is actually an extremely valuable side-benefit of the new `httpie` command.

The default `argparse` output is not helpful in this sense at all:

```bash
$ httpie
usage: httpie [-h] [--debug] [--traceback] {plugins} ...
httpie: error: please specify one of these: 'plugins'
```

We should also handle the following scenario. Here we could print the same as above, but also print the correct command (`http POST pie.dev/post hello=world`) so that the user can just copy/paste/re-run it with the correct command name this time:

```bash
$ httpie POST pie.dev/post hello=world
usage: httpie [-h] [--debug] [--traceback] {plugins} ...
httpie: error: argument action: invalid choice: 'POST' (choose from 'plugins')
```

This is slightly better:

```bash
$ httpie --help
usage: httpie [-h] [--debug] [--traceback] {plugins} ...

Managing interface for the HTTPie itself. <https://httpie.io/docs#manager>

Be aware that you might be looking for http/https commands for sending
HTTP requests. This command is only available for managing the HTTTPie
plugins and the configuration around it.

positional arguments:
  {plugins}

optional arguments:
  -h, --help
      show this help message and exit

  --debug
      Prints the exception traceback should one occur, as well as other
      information useful for debugging HTTPie itself and for reporting bugs.

  --traceback
      Prints the exception traceback should one occur.
```

isidentical:
@jakubroztocil I've implemented 2 different suggestions. First one is when you simply run `httpie`:

```bash
$ httpie                                                                                                                                            1,021s
usage: httpie [-h] [--debug] [--traceback] {plugins} ...
httpie: error: Please specify one of these: "plugins"

This command is for managing HTTPie plugins.
Perhaps you are looking for http/https commands, e.g:
    http POST pie.dev/post hello=world

```

It gives the actual description / error message, but also mentions about `http`/`https` commands (with a random example). The second is the case for `POST blahblah.com/dev A=B`:

```
 $ httpie POST pie.dev/post hello=world                                                                                                               576ms
usage: httpie [-h] [--debug] [--traceback] {plugins} ...
invalid choice: 'POST' (choose from 'plugins')

This command is for managing HTTPie plugins.
You probably meant one of the following: 
    http POST pie.dev/post hello=world
    https POST pie.dev/post hello=world
```

We try to parse the arguments as if it was parsed by `http`/`https` and if we succeed, we display that extra paragraph with the full examples.

jakubroztocil:
@isidentical thanks. I’ll play with it and tweak the copy a bit more. We got a failing test it seems: https://github.com/httpie/httpie/runs/4322055556?check_suite_focus=true#step:5:226

isidentical:
Yep, seems like the API I depend on is only available 3.9+. Trying to overcome that at the moment.

httpie/manager/__main__.py
```diff
@@ -0,0 +1,94 @@
...
+
+'''
+
+# noinspection PyStringFormat
+MSG_NAKED_INVOCATION = f'''\
+Please specify one of these: "plugins"
+
+{MSG_COMMAND_CONFUSION}
+
+'''.format(args='POST pie.dev/post hello=world')
```
Shivansh-007, line 28: This is still not very clear to me as auser who runs `httpie`, it should rather display the help message of `httpie` including the confusion paragraph at the starting.

httpie/manager/plugins.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def fail(
+        self,
+        command: str,
+        target: Optional[str] = None,
+        reason: Optional[str] = None
+    ) -> ExitStatus:
+        message = f'Can\'t {command}'
+        if target:
+            message += f' {target!r}'
```
Shivansh-007, line 51: why is `!r` needed? Isn't it already a string.

httpie/config.py
```diff
@@ -128,3 +128,7 @@ def __init__(self, directory: Union[str, Path] = DEFAULT_CONFIG_DIR):
     @property
     def default_options(self) -> list:
         return self['default_options']
+
+    @property
+    def plugins_dir(self) -> Path:
+        return Path(self.get('plugins_dir', self.directory / 'plugins')).resolve()
```
Shivansh-007, line 134: I don't think `~/.config/httpie/plugins` is the right direcctory to install the plugins into, they aren't really "configs", would better fit in `~/.local/share` or `~/.cache`

Discussion: https://discord.com/channels/725351238698270761/801485021016752208/914132385669472266

httpie/manager/plugins.py
```diff
@@ -0,0 +1,188 @@
...
+                    self.env.stderr.write(textwrap.indent('  ', stderr))
+
+                last_line = stderr.strip().splitlines()[-1]
+                severity, _, message = last_line.partition(': ')
+                if severity == 'ERROR':
+                    reason = message
+
+            return self.fail('install', ', '.join(targets), reason)
+
+    def _uninstall(self, target: str) -> Optional[ExitStatus]:
```
Shivansh-007, line 101: Only the main dep file gets uninstalled here, all the sub deps which it depends on are still present in the `site` directory. 

Eg: `httpie plugins uninstall httpie-nsof` would work perfectly as it has no dependencies (excluding `httpie`), whereas `httpie plugins uninstall httpie-oauth` wouldn't remove its depending deps such as `oauthlib`.

Shivansh-007 (commented):
Some comments, I am new to this lib (as a dev) so not really sure whether some of my comments are against the design.

docs/README.md
```diff
@@ -1915,6 +1915,59 @@ And since there’s neither data nor `EOF`, it will get stuck. So unless you’r
...
+<logs>
+Successfully installed httpie-plugin-1.0.2
+```
+
+#### `httpie plugins list`
+
+List all installed plugins.
+
+```bash
+$ httpie plugins list
```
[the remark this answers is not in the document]
isidentical: I don't think it makes sense for httpie, since we are not a fully-fledged package manager. There are only like ~50 plugins, with 10 of them actively developed. So I don't see any of the users dealing with this complexity, at this moment.

docs/README.md
```diff
@@ -1915,6 +1915,59 @@ And since there’s neither data nor `EOF`, it will get stuck. So unless you’r
...
+By default the plugins (and their missing dependencies) will be stored under the configuration directory,
+but this can be modified through `plugins_dir` variable on the config.
+
+#### `httpie plugins install`
+
+For installing plugins from [PyPI](https://pypi.org/) or from local paths, `httpie plugins install`
+can be used with multiple targets.
+
+```bash
+$ httpie plugins install httpie-plugin
```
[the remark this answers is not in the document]
isidentical: AFAIK you should be able to do that, `httpie-plugin==0.0.1`. We simply pass each item to pip, and pip accepts version specifiers.

httpie/manager/__main__.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+MSG_COMMAND_CONFUSION = '''
+This command is only for managing HTTPie plugins.
+To send a request, please use the http/https commands:
+
+  $ http {args}
+
+  $ https {args}
+
```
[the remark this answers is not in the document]
isidentical: I think @jakubroztocil added them for clarity.

httpie/manager/plugins.py
```diff
@@ -0,0 +1,188 @@
...
+        from httpie.plugins.registry import plugin_manager
+
+        known_plugins = defaultdict(list)
+
+        for entry_point in plugin_manager.iter_entry_points(self.dir):
+            ep_info = (entry_point.group, entry_point.name)
+            ep_name = get_dist_name(entry_point) or entry_point.module
+            known_plugins[ep_name].append(ep_info)
+
+        for plugin, entry_points in known_plugins.items():
```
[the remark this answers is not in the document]
isidentical: Isn't that the default behaviour of other listing tools, even `ls`? Though I guess we can print out `No plugins found.` message, WDYT @jakubroztocil?

docs/README.md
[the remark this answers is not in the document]
Shivansh-007: I didn't mean that level of compelexity, currently we output them as:
```bash
httpie_plugin (1.0.2)
  httpie_plugin (httpie.plugins.auth.v1)
httpie_converter (1.0.0)
  httpie_iterm_converter (httpie.plugins.converter.v1)
  httpie_konsole_konverter (httpie.plugins.converter.v1)
```
I suggesting to change that to:
```bash
httpie_plugin (1.0.2)
└── httpie_plugin (httpie.plugins.auth.v1)
httpie_converter (1.0.0)
├── httpie_iterm_converter (httpie.plugins.converter.v1)
└── httpie_konsole_konverter (httpie.plugins.converter.v1)
```

docs/README.md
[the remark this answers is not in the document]
Shivansh-007: Ah right 👍🏻

httpie/manager/__main__.py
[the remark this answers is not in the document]
Shivansh-007: Uhh, I don't see how that makes it clear.

httpie/manager/__main__.py
isidentical: What do you mean by that? We already print both the usage as well as the confusion paragraph.
```
 $ httpie
usage: httpie [-h] [--debug] [--traceback] {plugins} ...
httpie: error: Please specify one of these: "plugins"


This command is only for managing HTTPie plugins.
To send a request, please use the http/https commands:

  $ http POST pie.dev/post hello=world

  $ https POST pie.dev/post hello=world
```

httpie/manager/plugins.py
isidentical: `!r` calls `repr()` on the string, which makes it get printed with quotes. E.g:
```
>>> some_string = 's'
>>> print(f"{some_string}")
s
>>> print(f"{some_string!s}")
s
>>> print(f"{some_string!r}")
's'
```

httpie/config.py
isidentical: This was discussed on discord.

httpie/manager/__main__.py
Shivansh-007: I am saying we can directly just sent the output of `--help` if no arguments are specified because it sums the two messages (above) pretty well and it was more clear for me when looking at them.

httpie/manager/plugins.py
isidentical: This is simply how pip works, it is very problematic to backwards package dependency collection without storing extra metadata. It would bring a tremeounds amount of complexity, considering we would need to create a full-graph of dependency-relations and then remove them as well. As I mentioned, this is not a fully-fledged package manager but rather a simple overlay on top of pip.

Shivansh-007: Hmm, it's worth adding a comment over there as a `TODO` probably... I think there should be a simple way of doing this, will give it a shot when I get some time.

docs/README.md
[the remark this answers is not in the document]
Shivansh-007: (This isn't resolved).

httpie/manager/plugins.py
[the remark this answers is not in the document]
jakubroztocil: No output is clear enough

docs/README.md
```diff
@@ -1915,6 +1915,70 @@ And since there’s neither data nor `EOF`, it will get stuck. So unless you’r
...
+### `httpie plugins`
+
+`plugins` interface is a very simple plugin manager for installing, listing and uninstalling HTTPie plugins.
+
+> In the past `pip` was used to install/uninstall plugins, but on some environments (e.g brew installed
+packages) it wasn't working properly. The new interface is a very simple overlay on top of `pip` to allow
+plugin installations on every installation method.
+
+> By default the plugins (and their missing dependencies) will be stored under the configuration directory,
+but this can be modified through `plugins_dir` variable on the config.
```
jakubroztocil, line 1942: We probably should mention this under https://httpie.io/docs#configurable-options as well

docs/README.md
```diff
@@ -1915,6 +1915,70 @@ And since there’s neither data nor `EOF`, it will get stuck. So unless you’r
...
+can be used.
+
+```bash
+$ httpie plugins install httpie-plugin
+Installing httpie-plugin...
+Successfully installed httpie-plugin-1.0.2
+```
+
+> Tip: Generally HTTPie plugins start with `httpie-` prefix. Try searching for it on [PyPI](https://pypi.org/)
+> to find out all plugins from the community.
```
jakubroztocil, line 1956: This could link to `https://pypi.org/search/?q=httpie-`

httpie/manager/__main__.py
```diff
@@ -0,0 +1,64 @@
...
+    from httpie.cli.definition import parser as http_parser
+    from httpie.manager.cli import COMMANDS
+
+    # If the user already selected a top-level sub-command, never
+    # show the http/https version. E.g httpie plugins pie.dev/post
+    if (
+        len(args) >= 1
+        and args[0] in COMMANDS
+    ):
```
jakubroztocil, line 23: Is this Black’s doing? :D

jakubroztocil (commented):
Added a few last comments. Let’s marge it afterward, and we can always continue to polish in `master`  🚀

httpie/manager/plugins.py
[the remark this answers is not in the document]
Shivansh-007: `No plugins found.` is more clear than outputting nothing IMO.

httpie/manager/__main__.py
isidentical: Nop, this is my muscle memory automatically expanding 2 expressions in an if statement's test :D Black would've probably left them in the same line since it is within 88-chars. Fixed now!

PR #1203, opened 2021-11-10
Change `PyPi` to `PyPI`

doublevcodes:
This pull request changes the acronym for the Python Packaging Index from it’s previous spelling of PyPi to PyPI.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -58,7 +58,7 @@ Do not edit here, but in docs/installation/.
 
 ### Universal
 
-#### PyPi
+#### PyPI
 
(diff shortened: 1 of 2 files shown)


jakubroztocil (changes requested):
Thanks! But please see the comment a few lines above your change 🙃

janbrasna:
@doublevcodes In other words the right place to change this is in:

https://github.com/httpie/httpie/blob/06d9c14e7a0a8c5637a90931a5660401535a4d9b/docs/installation/methods.yml#L174

PR #1205, opened 2021-11-13
core: support custom request/response classes

isidentical:
While playing with a 3rd party extension (httpie-gssapi), I've noticed that it sometimes get broken due to the hard-coded `requests` classes. `requests` plugins like `requests-gssapi` might return different type of request/response classes like [requests_gssapi.gssapi_.SanitizedResponse](https://github.com/pythongssapi/requests-gssapi/blob/0402d14235783716635a1ea08bd0b1c63dc69272/requests_gssapi/gssapi_.py#L34-L60) which makes sense depending on the use case. But since some of the places within `httpie` hard codes this, it is impossible to get around of it (it uses `type(message)` instead of `isinstance()` etc. checks). 

This patch simply abstracts request/response inference to it's own unit, and simplifies this process by changing the default type resolution from bare `type()` checks to `isinstance()` checks so all classes of the `reqeusts.Response` and `requests.PreparedRequest` will be supported with this.

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -4,7 +4,7 @@ import json
 import sys
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Callable, Iterable, Union
+from typing import Callable, Iterable
 from urllib.parse import urlparse, urlunparse
 
 import requests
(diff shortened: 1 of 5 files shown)


jakubroztocil (changes requested):
Nice tweak! I’d prefix `message` with `requests` in all the new names, and move everything to `models.py`. There we already have related code.

PR #1208, opened 2021-11-19
Preserve individual headers with the same name on responses

### LINKED ISSUE #1207: Response headers sharing the same names get converted into a comma-separated list
## Checklist

- [x] I've searched for similar issues.
- [x] I'm using the latest version of HTTPie.

---

Make a request to URL that returns repeated headers lines. Here’s an example express server that does that:

```js
// server.js
const express = require('express')
const fs = require('fs')

const app = express()
const port = 4000

app.get('/', (req, res) => {
  res.set('x-foo', ['bar', 'baz'])
  res.send('Hello World!')
})

app.listen(port, () => {
  console.log(`Example app listening at http://localhost:${port}`)
})
```

The run `node server.js` and get the response:

```bash
http :4000
```

## Current result

```http
x-foo: bar, baz
```

## Expected result

```http
x-foo: bar
x-foo: baz
```

---

@isidentical we’ve fixed this for requests, so let’s fix it for the responses as well.

isidentical:
Resolves #1207

### DIFF
--- a/httpie/adapters.py
+++ b/httpie/adapters.py
@@ -1,13 +1,13 @@
-from httpie.cli.dicts import RequestHeadersDict
+from httpie.cli.dicts import HTTPHeadersDict
 from requests.adapters import HTTPAdapter
 
 
 class HTTPieHTTPAdapter(HTTPAdapter):
 
     def build_response(self, req, resp):
-        """Wrap the original headers with the `RequestHeadersDict`
+        """Wrap the original headers with the `HTTPHeadersDict`
         to preserve multiple headers that have the same name"""
 
         response = super().build_response(req, resp)
-        response.headers = RequestHeadersDict(getattr(resp, "headers", {}))
+        response.headers = HTTPHeadersDict(getattr(resp, "headers", {}))
         return response
--- a/httpie/adapters.py
+++ b/httpie/adapters.py
@@ -9,5 +9,5 @@ class HTTPieHTTPAdapter(HTTPAdapter):
         to preserve multiple headers that have the same name"""
 
         response = super().build_response(req, resp)
-        response.headers = HTTPHeadersDict(getattr(resp, "headers", {}))
+        response.headers = HTTPHeadersDict(getattr(resp, 'headers', {}))
         return response
--- a/httpie/adapters.py
+++ b/httpie/adapters.py
@@ -1,0 +1,13 @@
+from httpie.cli.dicts import RequestHeadersDict
+from requests.adapters import HTTPAdapter
+
+
+class HTTPieHTTPAdapter(HTTPAdapter):
+
+    def build_response(self, req, resp):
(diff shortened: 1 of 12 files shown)


httpie/adapters.py
```diff
@@ -0,0 +1,13 @@
...
+
+class HTTPieHTTPAdapter(HTTPAdapter):
+
+    def build_response(self, req, resp):
+        """Wrap the original headers with the `HTTPHeadersDict`
+        to preserve multiple headers that have the same name"""
+
+        response = super().build_response(req, resp)
+        response.headers = HTTPHeadersDict(getattr(resp, "headers", {}))
+        return response
```
isidentical, line 13: I am thinking about moving `httpie/ssl.py` into `httpie/adapters.py`, since it does make sense for all the adapters to be in the same place. WDYT @jakubroztocil?

tests/utils/http_server.py
```diff
@@ -0,0 +1,50 @@
...
+
+        return func(self)
+
+
+@TestHandler.handler('GET', '/headers')
+def get_headers(handler):
+    handler.send_response(200)
+    for key, value in handler.headers.items():
+        handler.send_header(key, value)
+    handler.send_header("Content-Length", 0)
```
jakubroztocil, line 35:
```suggestion
    handler.send_header('Content-Length', 0)
```

tests/utils/http_server.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.fixture(scope="function")
+def http_server():
+    """A custom HTTP server implementation for our tests, that is
+    built on top of the http.server module. Handy when we need to
+    deal with details which httpbin can not capture."""
+
+    server = HTTPServer(('localhost', 0), TestHandler)
+    thread = threading.Thread(target=server.serve_forever)
+    thread.start()
+    yield "{}:{}".format(*server.socket.getsockname())
```
jakubroztocil, line 48:
```suggestion
    yield '{}:{}'.format(*server.socket.getsockname())
```

httpie/adapters.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class HTTPieHTTPAdapter(HTTPAdapter):
+
+    def build_response(self, req, resp):
+        """Wrap the original headers with the `HTTPHeadersDict`
+        to preserve multiple headers that have the same name"""
+
+        response = super().build_response(req, resp)
+        response.headers = HTTPHeadersDict(getattr(resp, "headers", {}))
```
jakubroztocil, line 12:
```suggestion
        response.headers = HTTPHeadersDict(getattr(resp, 'headers', {}))
```

httpie/adapters.py
jakubroztocil: Sounds good 👍🏻

PR #1216, opened 2021-11-29
Implement Bearer Auth

### LINKED ISSUE #1215: Support Bearer Authentication
Bearer is a form of HTTP authentication through a header value (`Authorization: Bearer <token>`). It is pretty common and pretty well defined (by [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750)) so personally I'd say we should support it out of the box.

isidentical:
Resolves #1215

### DIFF
--- a/httpie/plugins/builtin.py
+++ b/httpie/plugins/builtin.py
@@ -34,6 +34,16 @@ class HTTPBasicAuth(requests.auth.HTTPBasicAuth):
         return f'Basic {token}'
 
 
+class HTTPBearerAuth(requests.auth.AuthBase):
+
+    def __init__(self, token: str) -> None:
+        self.token = token
+
+    def __call__(self, request: requests.PreparedRequest) -> requests.PreparedRequest:
+        request.headers['Authorization'] = f'Bearer {self.token}'
+        return request
+
+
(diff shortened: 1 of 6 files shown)


httpie/plugins/builtin.py
```diff
@@ -56,3 +66,14 @@ def get_auth(
...
+
+class BearerAuthPlugin(BuiltinAuthPlugin):
+    name = 'Bearer HTTP Auth'
+    auth_type = 'bearer'
+    netrc_parse = False
+    auth_parse = False
+
+    # noinspection PyMethodOverriding
+    def get_auth(self, **kwargs) -> requests.auth.HTTPDigestAuth:
+        return HTTPBearerAuth(self.raw_auth)
```
jakubroztocil, line 79: I wonder if it would make sense / be possible to enable `netrc_parse` / `prompt_password`

isidentical: `netrc` generally assumes the stuff is username/password (even though you can omit `username`, it just becomes an empty string), I am not sure if it is useful. If you feel like it, I can turn that back on with a test.

jakubroztocil:
🎉

PR #1219, opened 2021-11-30
Fix packit CI

isidentical:
Currently, the package in the Fedora repositories doesn't match the master, since we've added a new command. But packit doesn't know that and tries to build from the master with the 2.6.0 release spec file. This patch reconfigures packit to use our own locally hosted specfile.


darikankid (approved):

hroncok:
This can be reverted now.

PR #1224, opened 2021-12-02
Add nested JSON syntax to the HTTPie DSL

### LINKED ISSUE #78: Syntax for nested JSON
The simple examples work, but what about nested data?
In curl I do something like this:

-d \
'"credentials": {
    "username": "me",
    "key": "my-key"}
}'

How can I do this with httpie?

isidentical:
Implement nested JSON syntax to the HTTPie DSL (through JSON form notation). Resolves #78

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -714,6 +714,58 @@ The `:=`/`:=@` syntax is JSON-specific. You can switch your request to `--form` 
 and string, float, and number values will continue to be serialized (as string form values).
 Other JSON types, however, are not allowed with `--form` or `--multipart`.
 
+### Nested JSON fields
+
+Nested JSON fields can be set with both data field (`=`) and raw JSON field(`:=`) separators, which allows you to embed arbitrary JSON data into the resulting JSON object.
+HTTPie supports the [JSON form](https://www.w3.org/TR/html-json-forms/) syntax.
+
+```bash
+$ http PUT pie.dev/put \
+    'object=scalar' \                           # Object — blank key
+    'object[0]=array 1' \                       # Object — "0" key
+    'object[key]=key key' \                     # Object — "key" key
+    'array:=1' \                                # Array — first item
+    'array:=2' \                                # Array — second item
+    'array[]:=3' \                              # Array — append (third item)
+    'wow[such][deep][3][much][power][!]=Amaze'  # Nested object
+```
+
+```http
+PUT /person/1 HTTP/1.1
+Accept: application/json, */*;q=0.5
+Content-Type: application/json
+Host: pie.dev
+
+{
+    "array": [
+        1,
+        2,
+        3
+    ],
+    "object": {
+        "": "scalar",
+        "0": "array 1",
+        "key": "key key"
+    },
+    "wow": {
+        "such": {
+            "deep": [
+                null,
+                null,
+                null,
+                {
+                    "much": {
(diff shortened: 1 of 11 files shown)


docs/README.md
```diff
@@ -714,6 +714,246 @@ The `:=`/`:=@` syntax is JSON-specific. You can switch your request to `--form`
...
+    3
+]
+```
+
+Here is a slightly unified example
+
+```bash
+$ http --offline --print=B pie.dev/post name=python version:=3 \
+  date[year]:=2021 date[month]=December \
+  systems=Linux systems=Mac systems=Windows \
```
jakubroztocil, line 878: This actually removes the possibility to overwrite fields:

* `$ http --offline pie.dev a=1 a=2`
  * Old  → `{"a": 2}`
  * New → `{"a": [1, 2]}`

🤔

At some point I considered adding something like `--no-nested-json` to ease a bit the backward compatibility breakage. I.e., if my requests start breaking, I could just turn the nesting off without having to think too much.  What do you think?

httpie/cli/json_form.py
```diff
@@ -0,0 +1,150 @@
+'''
```
jakubroztocil, line 1: for these guys we use double quotes 😁

isidentical: Good to know 🎈

docs/README.md
isidentical: > At some point I considered adding something like --no-nested-json to ease a bit the backward compatibility breakage

I personally don't really prefer feature togglers, since these features are now part of the essential HTTPie DSL and it might just confuse people. Since depending on a CLI flag, the whole language processing changes. I'd prefer removing this behaviour, and only supporting array creation through `[]` and `[<index>]` over `--no-nested-json` flag. (If there aren't any more pitfalls).


The only bad side about this removal is that it would make us diverge from JSON form notation a bit more, which I think is acceptable considering the notation is not designed for CLI apps to use.

isidentical: Though now thinking a bit more, `=` already creates a list on `--form` mode. The same goes for `:` too. So I don't really know if overwriting should be incoherently supported in only JSON.

jakubroztocil: > The same goes for `:` too. So I don't really know if overwriting should be incoherently supported in only JSON.

For headers, the difference is that we have an [un-setting syntax](https://httpie.io/docs#empty-headers-and-header-un-setting) (just `Header:`). The idea is to allow you to overwrite a header specified in another way, e.g., in a session or the config file. 

For data fields, though, we can more strongly assume that the full payload is specified for each invocation, so the ability to overwrite previously specified fields is probably less important.

isidentical: > For headers, the difference is that we have an un-setting syntax (just Header:). The idea is to allow you to overwrite a header specified in another way, e.g., in a session or the config file.

Makes sense. Forgot about that.

> For data fields, though, we can more strongly assume that the full payload is specified for each invocation, so the ability to overwrite previously specified fields is probably less important.

So, how should we proceed? Leave it as is or should we consider other options (`--no-nested-json`/`not creating an array on path repetition`)?

jakubroztocil: I’d say leave it as it is (if you agree with my reasoning above).

isidentical: I do.

PR #1225, opened 2021-12-02
Formalize @ suffix for all operators

isidentical:
Resolves https://github.com/httpie/httpie/issues/964


docs/README.md
```diff
@@ -598,7 +598,14 @@ GET /../../etc/password HTTP/1.1
 
 There are a few different *request item* types that provide a convenient mechanism for specifying HTTP headers, simple JSON and form data, files, and URL parameters.
 
-They are key/value pairs specified after the URL. All have in common that they become part of the actual request that is sent and that their type is distinguished only by the separator used: `:`, `=`, `:=`, `==`, `@`, `=@`, `:=@` and `==@`. The ones with an `@` expect a file path as value.
+They are key/value pairs specified after the URL. All have in common that
+they become part of the actual request that is sent and that their type is
+distinguished only by the separator used: `:`, `=`, `:=`, `==`. If you want
```
jakubroztocil, line 603: There’s also the `field@file` syntax for file upload fields with `--form`:

```bash
$ http --offline pie.dev/post  file@/etc/hosts
```

docs/README.md
```diff
@@ -598,7 +598,14 @@ GET /../../etc/password HTTP/1.1
...
+to use a file's contents with these operators, you can simply add an `AT` (`@`)
+operator as the suffix. Which is going to make them  read the file's contents,
+and use it as the value for the specified request item (`:` and `==` will strip
+any trailing newlines automatically). The `AT` (`@`) operator itself can also be
+individually used to embed arbitrary data directly into the request's body.
```
jakubroztocil, line 608: Here we should link to https://httpie.io/docs#request-data-from-a-filename

jakubroztocil, line 608: This paragraph introduces a fundamental concept and currently, it feels rather dense. Please continue working on it to make it more digestible for people just getting started with HTTPie. And the table below is inconsistent as it shows the `@` suffix for one of the types only. 

Also, the lack of examples makes it very abstract. I think we should add a comprehensive example similar to the JSON one:

```
$ http PUT pie.dev/put \
    name=John \                        # String (default)
    age:=29 \                          # Raw JSON — Number
    married:=false \                   # Raw JSON — Boolean
    hobbies:='["http", "pies"]' \      # Raw JSON — Array
    favorite:='{"tool": "HTTPie"}' \   # Raw JSON — Object
    bookmarks:=@files/data.json \      # Embed JSON file
    description=@files/text.txt        # Embed text file
```

Maybe a bigger refactoring is needed.

jakubroztocil: Now that we’br had added the `@` suffix to headers and params, we should mention them in their respective sections as well (with a backlink to the request items section).

jakubroztocil (changes requested):
Code looks good 👍🏻 
Docs need some tweaks 📓

isidentical:
@jakubroztocil please re-review, I've tried to address your suggestions and make it a bit more simple.

docs/README.md
```diff
@@ -596,21 +602,48 @@ GET /../../etc/password HTTP/1.1
...
+Each *request item* is simply a key/value pair separated with the following
+characters: `:` (headers), `=` (data field, e.g JSON, Form), `:=` (raw data field)
+`==` (query parameters), `@` (file upload).
 
-They are key/value pairs specified after the URL. All have in common that they become part of the actual request that is sent and that their type is distinguished only by the separator used: `:`, `=`, `:=`, `==`, `@`, `=@`, `:=@` and `==@`. The ones with an `@` expect a file path as value.
+```bash
+$ http PUT pie.dev/put \
+    X-Date:today \                     # Header
+    token==secret \                    # Query parameter
+    name=John \                        # String (default)
```
jakubroztocil, line 618: `String (default)` doesn’t explain what it does

jakubroztocil: The space before the comments don’t need to be to so big

isidentical: Perhaps
```suggestion
    name=John \                        # String Data (JSON or Form)
```

docs/README.md
```diff
@@ -596,21 +602,48 @@ GET /../../etc/password HTTP/1.1
...
+
+Using file contents as values for specific fields is a very common use case, which can be achieved through adding the `@` suffix to
+the operators above. For example instead of using a static string as the value for some header, you can use `:@` operator
+to pass the desired value from a file.
+
+```bash
+$ http POST pie.dev/post \
+    X-Data:@files/text.txt             # Read a header from a file
+    token==@files/text.txt             # Read a query parameter from a file
+    name=@files/text.txt               # Read a string from a file
```
jakubroztocil, line 643: Again, "string" says nothing about how it’s used

docs/README.md
```diff
@@ -871,6 +904,14 @@ Host: <taken-from-URL>
 
 Any of these can be overwritten and some of them unset (see below).
 
+### Reading headers from a file
+
+You can read headers from a file by using the `:@` operator. This would also effectively strip the newlines from the end.
```
jakubroztocil, line 909: Back link to the section dedicated to the file operators

docs/README.md
```diff
@@ -518,6 +518,12 @@ $ http https://api.github.com/search/repositories q==httpie per_page==1
 GET /search/repositories?q=httpie&per_page=1 HTTP/1.1
 ```
 
+You can even retrieve the `value` from a file by using the `param==@file` syntax. This would also effectively strip the newlines from the end.
```
jakubroztocil, line 521: Back link to the section dedicated to the file operators

docs/README.md
```diff
@@ -871,6 +904,14 @@ Host: <taken-from-URL>
 
 Any of these can be overwritten and some of them unset (see below).
 
+### Reading headers from a file
+
+You can read headers from a file by using the `:@` operator. This would also effectively strip the newlines from the end.
+
+```bash
+$ http pie.dev/headers :@files/text.txt
```
jakubroztocil, line 912: Missing header name

jakubroztocil (changes requested):

docs/README.md
jakubroztocil: The table below describes the concepts so labels should be consistent with that. This is a data field.

PR #1226, opened 2021-12-02
Automatically enable --stream on server sent events

### LINKED ISSUE #376: Server-sent events explicitly requires --stream
If I have a [server-sent event](https://en.wikipedia.org/wiki/Server-sent_events) source and try to simply do

```
http get http://localhost:1729/stream/all
```

I get output that looks something like this:

```
HTTP/1.1 200 OK
Cache-Control: no-cache
Content-Type: text/event-stream
Date: Sat, 29 Aug 2015 12:18:00 GMT
Server: TornadoServer/4.2.1
Transfer-Encoding: chunked

```

followed by none of the streamed data. This can be worked around by setting the `--stream` flag. However, better behavior would be to automatically enable streaming support when `Content-Type` is `text/event-stream`.

isidentical:
Resolves #376.

### DIFF
--- a/tests/test_uploads.py
+++ b/tests/test_uploads.py
@@ -9,7 +9,7 @@ from .utils import (
     MockEnvironment, StdinBytesIO, http,
     HTTP_OK,
 )
-from .fixtures import FILE_PATH_ARG, FILE_PATH, FILE_CONTENT
+from .fixtures import FILE_PATH_ARG, FILE_PATH, FILE_CONTENT, RAW_FILE_CONTENT
 
 
 def test_chunked_json(httpbin_with_chunked_support):
@@ -49,7 +49,8 @@ def test_chunked_stdin(httpbin_with_chunked_support):
     )
     assert HTTP_OK in r
     assert 'Transfer-Encoding: chunked' in r
-    assert r.count(FILE_CONTENT) == 2
+    # One original (request) + one encoded (httpbin's response, ascii)
+    assert r.count(FILE_CONTENT) + r.count(RAW_FILE_CONTENT) == 2
 
 
 def test_chunked_stdin_multiple_chunks(httpbin_with_chunked_support):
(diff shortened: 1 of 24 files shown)


tests/test_uploads.py
```diff
@@ -49,7 +49,8 @@ def test_chunked_stdin(httpbin_with_chunked_support):
     )
     assert HTTP_OK in r
     assert 'Transfer-Encoding: chunked' in r
-    assert r.count(FILE_CONTENT) == 2
+    # One original (request) + one encoded (httpbin's response, ascii)
+    assert r.count(FILE_CONTENT) + r.count(RAW_FILE_CONTENT) == 2
```
isidentical, line 53: The reason for these changes is that, httpbin mirrors the headers when it is responding. Since it is also mirroring the `Transfer-Encoding`, we treat this response as if it was chunked and use different delimiting techniques. So because of that, we stop formatting it as JSON, and not decode the actual unicode symbols (since that happens on formatting). Which I think is a plausible behavior, considering in this case, the server is the wrong side.

isidentical:
Tests are failing on 3.7/3.6, probably due to some mocking behavior changes. Will look at them on the morning.

jakubroztocil:
Why is it enabled for all chunked responses? This really only makes sense for server-sent events, where each line can be considered a body.   https://httpie.io/docs#disabling-buffering

isidentical:
> Why is it enabled for all chunked responses? This really only makes sense for server-sent events, where each line can be considered a body.

Ah, alright! Will do that then, thanks for the correction. I thought you initially meant enabling this for all chunked responses, since the quote in [this](https://github.com/httpie/httpie/issues/376#issuecomment-393269549) implied that. Will fix it!

mivade:
Woah, this is great! I'm not using SSE as much as I was when I filed #376 but this is really nice to see.

PR #1227, opened 2021-12-03
Don't inconsistently add XML declarations

### LINKED ISSUE #1156: Do not append missing XML declaration in formatted responses

isidentical:
Resolves #1156


httpie/output/formatters/xml.py
```diff
@@ -8,13 +8,27 @@
     from xml.dom.minidom import Document
 
 
+XML_OPEN_TAG = '<?xml'
+XML_CLOSE_TAG = '?>'
```
jakubroztocil, line 12: Pedantic: Would be nice to `DECL` in the name, and not have `TAG` in there as it can be confused with elements tags.

httpie/output/formatters/xml.py
```diff
@@ -26,9 +40,18 @@ def pretty_xml(document: 'Document',
     if standalone is not None and sys.version_info >= (3, 9):
         kwargs['standalone'] = standalone
     body = document.toprettyxml(**kwargs).decode(kwargs['encoding'])
+    lines = [line for line in body.splitlines() if line.strip()]
+
+    # Remove the declaration
+    if parse_declaration(lines[0]):
```
jakubroztocil, line 46: Will there always be `lines[0]`?

httpie/output/formatters/xml.py
```diff
@@ -26,9 +40,18 @@ def pretty_xml(document: 'Document',
     if standalone is not None and sys.version_info >= (3, 9):
```
jakubroztocil, line 40: Why do we deal with `standalone`? I think it should have no impact on our use case, and we should treat all XML docs as standalone (to avoid any potential loading of linked external stuff).

jakubroztocil, line 40: If we indeed need this, then it would be nice to move the version check to `compat.py`.

jakubroztocil (commented):
I think we should also include a test that verifies the behavior we’re correcting (i.e., that an unwanted XML declaration is indeed added by the library).

---

I just noticed that the parser also collapses empty tags:

```bash
$ echo '<a></a>' | http --offline pie.dev/post content-type:application/xml
```

```xml
<?xml version="1.0" encoding="utf-8"?>
<a/>
```

And I assume that might not be the last unwanted alteration. Not sure if we should spend too much time trying to fix this. 

(I was against reintroducing the XML formatter but was assured that this one has none the issues of the original one, but it seems like we’re in a similar situation.)

httpie/output/formatters/xml.py
isidentical: > Will there always be lines[0]?

That is a good question, and from my understanding the answer is yes. Everytime we convert the document to a string, the root of it will provide unconditional declaration (https://github.com/python/cpython/blob/851811f5772c43f72f445e2ce1ac3ea9da951ae3/Lib/xml/dom/minidom.py#L1823).

httpie/output/formatters/xml.py
isidentical: I have not touched/interacted with this part, and I don't really know. We pass it depending on whether the initial declaration is `standalone` or not.

isidentical:
> I think we should also include a test that verifies the behaviour we’re correcting (i.e., that an unwanted XML declaration is indeed added by the library).

We already have such tests (we had tests that did not have a declaration in their raw version, but have it on their formatted. So I simply fixed them).

> And I assume that might not be the last unwanted alteration. Not sure if we should spend too much time trying to fix this.

It is due to Python's parser is not really caring about roundtripability. I've looked around on PyPI, but I have not yet found an XML parser that claims to have a full-roundtripability feature. If that is what we are looking at, we could drop this or explicitly mention about the formatters are only looking for semantical equivalence.

httpie/output/formatters/xml.py
jakubroztocil: > > I think we should also include a test that verifies the behavior we’re correcting (i.e., that an unwanted XML declaration is indeed added by the library).
> 
> We already have such tests (we had tests that did not have a declaration in their raw version, but have it on their formatted. So I simply fixed them).

What I mean is that we implicitly rely on the library auto-adding the declaration. For example, we assume there’ll always be `lines[0]`. 

Let’s add an explicit test case verifying this assumption still holds (something like `assert pretty('<x/>') == '<?xml …?><x/>'`).

 
> explicitly mention about the formatters are only support for semantical equivalence.

Let’s do that.

jakubroztocil:
> > I think we should also include a test that verifies the behavior we’re correcting (i.e., that an unwanted XML declaration is indeed added by the library).
> 
> We already have such tests (we had tests that did not have a declaration in their raw version, but have it on their formatted. So I simply fixed them).

What I mean is that we implicitly rely on the library auto-adding the declaration. For example, we assume there’ll always be `lines[0]`. 

Let’s add an explicit test case verifying this assumption still holds (something like `assert pretty('<x/>') == '<?xml …?><x/>'`).

 
> explicitly mention about the formatters are only support for semantical equivalence.

Let’s do that.

httpie/output/formatters/xml.py
jakubroztocil: Could you please still clean it up even though it’s older code?

httpie/output/formatters/xml.py
```diff
@@ -26,9 +40,18 @@ def pretty_xml(document: 'Document',
...
+    if parse_declaration(lines[0]):
+        lines.pop(0)
+
+    # And replace it with the original, if it is present.
+    if declaration:
+        lines.insert(0, declaration)
 
     # Remove blank lines automatically added by `toprettyxml()`.
-    return '\n'.join(line for line in body.splitlines() if line.strip())
+    return '\n'.join(lines)
```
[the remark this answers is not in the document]
jakubroztocil: I don’t understand. This change undoes `# Remove blank lines automatically added by toprettyxml().`. Why is that?

isidentical:
> What I mean is that we implicitly rely on the library auto-adding the declaration. For example, we assume there’ll always be lines[0].

I see. I've changed it now to ensure there is at least one line and the first line is some sort of a declaration. If that is the case, we either get rid of it or replace it with the original version (if there are any).

> Let’s do that.

Sure!

httpie/output/formatters/xml.py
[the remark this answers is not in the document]
isidentical: It actually doesn't, it just moves it upwards. I didn't notice the comment, will move it as well.

httpie/output/formatters/xml.py
isidentical: Sure thing.

PR #1232, opened 2021-12-07
Add initial benchmarking infrastructure

isidentical:


isidentical:
Here is a real example: https://github.com/httpie/httpie/pull/1233#issuecomment-987864883

jakubroztocil (changes requested):
Good stuff.

- [ ] it would be good to explain the system (including the labeling) —perhaps add it to `CONTRIBUTING.md` in addition to the PR’s description
- [ ] the output should be super easy to skim: currently, it’s not super clear what each test does and it takes some effort to understand  what the impact of the PR is
- [ ] here we could use some more docs in the code as well (I’m generally for self-explanatory rather than commented code, but here they’d add value)

isidentical:
@jakubroztocil how about this?
https://github.com/httpie/httpie/pull/1233#issuecomment-989733988
image

isidentical:
Another example;
image

PR #1238, opened 2021-12-10
Implement new `pie` and `pie-light` styles

### LINKED ISSUE #1237: Create a style that matches with the HTTPie for Web/Desktop
[HTTPie for Web and Desktop](https://httpie.io/beta) uses a very cool colour scheme, and we should probably add it as a custom style to the CLI.

isidentical:
Resolves #1237.

Demos:
[asciicast](https://asciinema.org/a/GEJzIPv1mVwP5d8Q3oCM4QjSr)
[asciicast](https://asciinema.org/a/0PzxZaOjEsr41I3wzQmjOMmGv)


httpie/output/formatters/colors.py
```diff
@@ -237,3 +268,142 @@ class Solarized256Style(pygments.style.Style):
...
+        ('universal', '600'),
+        ('dark', '500')
+    ]:
+        yield type(
+            name.format(mode=mode.title()),
+            (pygments.style.Style,),
+            {
+                'styles': format_style(raw_styles, shade)
+            }
+        )
```
isidentical, line 308: We probably should use a more robust methodology in the future (for -100 -200 -300 etc). Something like this
```py
for shade in range(100, 800, 100):
    name = SHADE_NAMES.get(shade, f"pie-{shade}")
    for style_type, style_map in [("Header", HEADER_COLORS), ("Body", BODY_COLORS)]:
         create_styles()
```

though I plan to omit this for this release, and implement this after this gets merged (if there are no objections).

docs/README.md
```diff
@@ -1717,13 +1717,16 @@ Syntax highlighting is applied to HTTP headers and bodies (where it makes sense)
...
-|  `fruity` | A bold, colorful scheme. Enable with `--style=fruity`                                                                               |
-|         … | See `$ http --help` for all the possible `--style` values                                                                           |
+|     Style  | Description                                                                                                                          |
+| ---------: | ------------------------------------------------------------------------------------------------------------------------------------ |
+|     `auto` | Follows your terminal ANSI color styles. This is the default style used by HTTPie                                                    |
+|  `default` | Default styles of the underlying Pygments library. Not actually used by default by HTTPie. You can enable it with `--style=default`  |
+| `pie-dark` | HTTPie's original brand style. Also used in the HTTPie for Web and Desktop.                                                          |
```
jakubroztocil, line 1724:
```suggestion
| `pie-dark` | HTTPie’s original brand style. Also used in [HTTPie for Web and Desktop](https://httpie.io/product).                                                          |
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for _receiving_ multiple HTTP headers lines with the same name. ([#1207](https://github.com/httpie/httpie/issues/1207))
 - Added support for basic JSON types on `--form`/`--multipart` when using JSON only operators (`:=`/`:=@`). ([#1212](https://github.com/httpie/httpie/issues/1212))
 - Added support for automatically enabling `--stream` when `Content-Type` is `text/event-stream`. ([#376](https://github.com/httpie/httpie/issues/376))
+- Added new `pie-dark`/`pie-light` (and `pie`) styles that match with the [HTTPie for Web and Desktop](https://httpie.io/app). ([#1237](https://github.com/httpie/httpie/issues/1237))
```
jakubroztocil, line 17:
```suggestion
- Added new `pie-dark`/`pie-light` (and `pie`) styles that match with [HTTPie for Web and Desktop](https://httpie.io/product). ([#1237](https://github.com/httpie/httpie/issues/1237))
```

docs/README.md
```diff
@@ -1717,13 +1717,16 @@ Syntax highlighting is applied to HTTP headers and bodies (where it makes sense)
...
+|`pie-light` | Like `pie-dark`, but for terminals with light background colors.                                                                     |
+|      `pie` | A general version of `pie-dark` and `pie-light` themes that can work with any terminal background                                    |
```
jakubroztocil, line 1726:
```suggestion
|      `pie` | A generic version of `pie-dark` and `pie-light` themes that can work with any terminal background. Its universality requires compromises in terms of legibility, but it’s useful if you frequently switch your terminal between dark and light backgrounds.                                    |
```

httpie/output/formatters/colors.py
jakubroztocil: 👍🏻

jakubroztocil, line 308: These are tied to the Pie themes, so I’d either reflect that in their names or move them to `palette`.

httpie/output/ui/palette.py
```diff
@@ -0,0 +1,138 @@
+# Copy the brand palette
+
+COLOR_PALETTE = {
```
jakubroztocil, line 3: `PIE_COLOR_PALETTE`

httpie/output/formatters/colors.py
```diff
@@ -15,22 +16,32 @@
...
 DEFAULT_STYLE = AUTO_STYLE
-SOLARIZED_STYLE = 'solarized'  # Bundled here
+
+# Bundled here
+SOLARIZED_STYLE = 'solarized'
+PIE_STYLES = {
+    'pie',
+    'pie-light',
+    'pie-dark'
+}
```
jakubroztocil, line 34: Looks like an enum/constants candidate (we refer to it from multiple places). Maybe it could also go to `palette`.

jakubroztocil (changes requested):
Good stuff 💅 A few remaining tweaks.

httpie/output/ui/palette.py
isidentical: Do we need this prefix? I assumed these were going to be the default colors we use after this PR (e.g on progress bars, on regular colored outputs etc.)

jakubroztocil: Good question. I thought we’d always re-use the effective `--style` even for the UI, because if I use the terminal scheme, for example, then I’d like to see those colors everywhere. In that case, giving it the `PIE_` prefix would make sense.

isidentical: I especially put this into a separate package, `httpie.output.ui`, since I plan to (at least feel like) this is going to be the place where we would manage our rich UI (progress bars, themes, highlighting, console, etc.), and those probably will use this theme. If you feel strong about it, I can add the `PIE_` prefix though but personally I'd find this very compelling (since the fully qualified name of this is `httpie.output.ui.COLOR_PLAETTE`, so it is not something arbitrary).

jakubroztocil: OK. Although we should discuss this:

> I thought we’d always re-use the effective `--style` even for the UI, because if I use the terminal scheme, for example, then I’d like to see those colors everywhere.

Issue #1098, opened 2021-06-28
Prompted passwords are not properly persisted into sessions


blyxxyz:
**What are the steps to reproduce the problem?**

1. While using a session, supply only a username to `--auth` so the password is prompted. e.g. `http --offline --session ./sess.json --auth user :`
2. Send a new request using the session without explicit authentication, e.g. `http --offline --session ./sess.json :`

**What is the expected result?**

One of the following:
- The password I entered last time is sent
- I have to enter my password again

(Which of these would be desirable? We're implementing sessions in ducaale/xh#125 and would like to stay compatible)

**What happens instead?**

In the new request, the password is sent as `None`.

**Example**

```console
$ http --offline --session ./sess.json --auth user :
http: password for user@localhost: pass
GET / HTTP/1.1
[...]
Authorization: Basic dXNlcjpwYXNz
[...]

$ echo dXNlcjpwYXNz | base64 -d
user:pass

$ http --offline --session ./sess.json :
GET / HTTP/1.1
[...]
Authorization: Basic dXNlcjpOb25l
[...]

$ echo dXNlcjpOb25l | base64 -d
user:None

$ cat sess.json
[...]
    "auth": {
        "raw_auth": "user",
        "type": "basic"
    },
[...]
```

isidentical:
Fixed by #1239

PR #1239, opened 2021-12-11
Solution for issue 1098 - prompted passwords are not properly persisted into sessions

sebastianczech:
Pull request is proposition of solution for the issue #1098 , where prompted passwords are not properly persisted into sessions. Below is an example how it works after applying solution: 

[asciicast](https://asciinema.org/a/twmDRfaXgTS3nRNDH1INsq0pp)

This is my first contribution into open source, so all feedback is more than welcome ;)

### DIFF
--- a/httpie/cli/argtypes.py
+++ b/httpie/cli/argtypes.py
@@ -134,7 +134,7 @@ class AuthCredentials(KeyValueArg):
     """Represents parsed credentials."""
 
     def has_password(self) -> bool:
-        return bool(self.value)
+        return self.value is not None
 
     def prompt_password(self, host: str):
         prompt_text = f'http: password for {self.key}@{host}: '
--- a/httpie/cli/argtypes.py
+++ b/httpie/cli/argtypes.py
@@ -134,7 +134,7 @@ class AuthCredentials(KeyValueArg):
     """Represents parsed credentials."""
 
     def has_password(self) -> bool:
-        return self.value
+        return bool(self.value)
 
     def prompt_password(self, host: str):
         prompt_text = f'http: password for {self.key}@{host}: '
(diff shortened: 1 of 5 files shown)


httpie/cli/argtypes.py
```diff
@@ -134,7 +134,7 @@ class AuthCredentials(KeyValueArg):
     """Represents parsed credentials."""
 
     def has_password(self) -> bool:
-        return self.value is not None
+        return self.value
```
isidentical, line 137: I don't understand this change, why are we returning the password in `has_password`. `value` can be either a string (which is not a bool, check the function's signature) or `None`.

isidentical (changes requested):
The solution seems a bit localized, see my comments. Other than that; I am not really a big fan of storing raw passwords locally, so I would personally go with re-asking it rather than storing it as plain text. CC: @jakubroztocil on the product decision

httpie/cli/argtypes.py
sebastianczech: My mistake, I have fixed by changing it into ``bool(self.value)``. In some cases I saw, that password is ``None``, but sometimes it's empty string ``''``.

httpie/plugins/builtin.py
```diff
@@ -51,6 +51,7 @@ class BasicAuthPlugin(BuiltinAuthPlugin):
 
     # noinspection PyMethodOverriding
     def get_auth(self, username: str, password: str) -> HTTPBasicAuth:
+        self.raw_auth = username + ":" + password
```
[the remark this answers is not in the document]
sebastianczech: You are right. I removed from authentication plugin and move it into ``argparser.py`` , function ``_process_auth``. Is this approach acceptable ?

Amirjokr (approved):

isidentical (approved):
Thanks a lot for the PR @sebastianczech. I've noticed that the change you have made on the `has_password` broke the use case of empty passwords, so I manually fixed it and pushed it. The rest LGTM. Amazing first contribution 🎉

sebastianczech:
Thank you @isidentical for review, fixing `has_password` and merging changes

PR #1243, opened 2021-12-15
Optimize encoding detection

isidentical:
In the past we detected encoding once per chunk, which is very inefficient. Since if a chunk has a certain encoding, then the next one is going to be the same. This optimizes `--pretty={all, color, format}` by up to %35 (or even more if the response size is huge). 

Before:
```
Summary
  'http --print=HBhb --stream --ignore-stdin --pretty=none  pie.dev/stream/1000' ran
    1.45 ± 0.11 times faster than 'http --print=HBhb --stream --ignore-stdin --pretty=colors  pie.dev/stream/1000'
    1.47 ± 0.11 times faster than 'http --print=HBhb --stream --ignore-stdin --pretty=format  pie.dev/stream/1000'
    1.47 ± 0.08 times faster than 'http --print=HBhb --stream --ignore-stdin --pretty=all  pie.dev/stream/1000'
```

After:
```
Summary
  'http --print=HBhb --stream --ignore-stdin --pretty=format  pie.dev/stream/1000' ran
    1.02 ± 0.05 times faster than 'http --print=HBhb --stream --ignore-stdin --pretty=all  pie.dev/stream/1000'
    1.09 ± 0.10 times faster than 'http --print=HBhb --stream --ignore-stdin --pretty=none  pie.dev/stream/1000'
    1.11 ± 0.08 times faster than 'http --print=HBhb --stream --ignore-stdin --pretty=colors  pie.dev/stream/1000'
```

### DIFF
--- a/extras/profiling/benchmarks.py
+++ b/extras/profiling/benchmarks.py
@@ -178,9 +178,8 @@ for pretty in ['all', 'none']:
         f'`http --pretty={pretty} pie.dev/stream/1000`',
         [
             '--print=HBhb',
-            '--stream',
             f'--pretty={pretty}',
-            'httpbin.org/stream/100'
+            'httpbin.org/stream/1000'
         ]
     )
 DownloadRunner('download', '`http --download :/big_file.txt` (3GB)', '3G')
--- a/httpie/encoding.py
+++ b/httpie/encoding.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Union, Tuple
 
 from charset_normalizer import from_bytes
 from charset_normalizer.constant import TOO_SMALL_SEQUENCE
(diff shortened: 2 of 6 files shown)


isidentical:
Automatic encoding detecting through `charset` parameter from `content-type` is still there (it was an old feature, I did not add it).  This is only for optimizing when the response headers do not specify a `charset`.

jakubroztocil (commented):
What an elegant fix!

However, I think the detection is currently done on every chunk because if the 1st line is just whitespace or ASCII-compatible start of a non-ASCII document, for example, then the detected encoding is unreliable. 

Perhaps we need some middle ground:

- Original: fully pessimistic
- New: fully optimistic
- Middle ground: stop detecting after a certain confidence interval has been reached

What do you think?

(We should probably not spend too much time on this, though. I also think `--stream` use cases might be more IO-bound.)

PR #1244, opened 2021-12-15
Strip out extra variables from the actual mime type

### LINKED ISSUE #1242: No syntax highlighting possibly for content-types with charset
```bash
$ https pie.dev

```

isidentical:
Resolves #1242. We used to pass the whole `Content-Type` header as the mime (including the stuff like `;charset=utf-8` and other variables), with this patch we now only pass the actual mime type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for automatically enabling `--stream` when `Content-Type` is `text/event-stream`. ([#376](https://github.com/httpie/httpie/issues/376))
 - Broken plugins will no longer crash the whole application. ([#1204](https://github.com/httpie/httpie/issues/1204))
 - Fixed auto addition of XML declaration to every formatted XML response. ([#1156](https://github.com/httpie/httpie/issues/1156))
+- Fixed highlighting when `Content-Type` specifies an additional charset. ([#1242](https://github.com/httpie/httpie/issues/1242))
 
 ## [2.6.0](https://github.com/httpie/httpie/compare/2.5.0...2.6.0) (2021-10-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for automatically enabling `--stream` when `Content-Type` is `text/event-stream`. ([#376](https://github.com/httpie/httpie/issues/376))
 - Broken plugins will no longer crash the whole application. ([#1204](https://github.com/httpie/httpie/issues/1204))
 - Fixed auto addition of XML declaration to every formatted XML response. ([#1156](https://github.com/httpie/httpie/issues/1156))
-- Fixed highlighting when `Content-Type` specifies an additional charset. ([#1242](https://github.com/httpie/httpie/issues/1242))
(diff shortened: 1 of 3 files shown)


tests/utils/http_server.py
```diff
@@ -52,6 +52,18 @@ def chunked_drip(handler):
...
+@TestHandler.handler('GET', '/html')
+def basic_html_page(handler):
+    content = b"\x00<some></some>"
+
+    handler.send_response(200)
+    handler.send_header('Content-Type', 'text/html; charset=utf-8')
+    handler.send_header('Content-Length', len(content))
+    handler.end_headers()
+
+    handler.wfile.write(content)
```
jakubroztocil, line 64: You could just use the `httpbin` fixture (`httpbin + '/html'`).

```
$ http pie.dev/html
HTTP/1.1 200 OK
CF-Cache-Status: DYNAMIC
CF-RAY: 6be0bf830dce2780-PRG
Connection: keep-alive
Content-Encoding: gzip
Content-Type: text/html; charset=utf-8
```

isidentical: I wanted to ensure the mime type gets matched by converters but unfortunately I was not able to modify httpbin's html endpoint's data (converts only get activated when the data includes `\x00`).

jakubroztocil: I think you could simply check if there’s some color in there:

```python
from .utils import http, COLOR


def test_html(httpbin):
    r = http('--body', httpbin + '/html')
    assert COLOR in r
````

Comment (jakubroztocil) [reply tests/utils/http_server.py:64]: There’s usually a way to test pretty anything just against `httpbin` :) None of those mocking tests have been added by me.

Here, you might actually also need to provide a custom `Environment` to enable colorized output (`stdout_isatty=True`)

File: tests/test_output.py [hunk 2]
```diff
@@ -215,6 +215,18 @@ def test_get_lexer_not_found(self):
...
+@pytest.mark.parametrize("endpoint", [
+    "/encoding/utf8",
+    "/html",
+    "/json",
+    "/xml",
+])
+def test_ensure_contents_colored(httpbin, endpoint):
+    env = MockEnvironment(colors=256)
+    r = http('GET', httpbin + endpoint, env=env)
+    assert COLOR in r
```
jakubroztocil, line 227: You need to ensure headers are not included in the output for the `COLOR in r` check to be meaningful (e.g., [add `--body`](https://github.com/httpie/httpie/pull/1244#discussion_r769772969))

jakubroztocil (changes requested):
We also want to mention this in the changelog I believe

tests/test_output.py
isidentical: Right, makes sense.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for automatically enabling `--stream` when `Content-Type` is `text/event-stream`. ([#376](https://github.com/httpie/httpie/issues/376))
 - Broken plugins will no longer crash the whole application. ([#1204](https://github.com/httpie/httpie/issues/1204))
 - Fixed auto addition of XML declaration to every formatted XML response. ([#1156](https://github.com/httpie/httpie/issues/1156))
+- Fixed highlighting when `Content-Type` specifies an additional charset. ([#1242](https://github.com/httpie/httpie/issues/1242))
```
jakubroztocil, line 19:
```suggestion
- Fixed highlighting when `Content-Type` specifies `charset`. ([#1242](https://github.com/httpie/httpie/issues/1242))
```

PR #1254, opened 2021-12-27
Fix --raw with --chunked

### LINKED ISSUE #1253: --chunked with --raw raises an internal error
```console
$ http --chunked pie.dev/post --raw '{"a": 1}'                             
http: error: AttributeError: 'bytes' object has no attribute 'encode'
```

isidentical:
Resolves #1253. This is also a rewrite of the core uploading logic.

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -273,6 +273,26 @@ def make_send_kwargs_mergeable_from_env(args: argparse.Namespace) -> dict:
     }
 
 
+def as_json(data) -> str:
+    """Convert the given JSON dict into the actual request body."""
+
+    # Propagate the top-level list if there is only one
+    # item in the object, with an en empty key.
+    if len(data) == 1:
+        [(key, value)] = data.items()
+        if key == '' and isinstance(value, list):
+            data = value
+
+    if data:
+        data = json.dumps(data)
+    else:
+        # We need to set data to an empty string to prevent requests
+        # from assigning an empty list to `response.request.data`.
+        data = ''
+
+    return data
+
+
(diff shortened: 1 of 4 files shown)


httpie/client.py
```diff
@@ -273,6 +273,26 @@ def make_send_kwargs_mergeable_from_env(args: argparse.Namespace) -> dict:
     }
 
 
+def as_json(data) -> str:
+    """Convert the given JSON dict into the actual request body."""
```
jakubroztocil, line 277: Let’s convert the comment into a nicer functions name and annotate `data`.

httpie/uploads.py
```diff
@@ -36,69 +42,101 @@ def __iter__(self) -> Iterable[Union[str, bytes]]:
...
+    if isinstance(data, str):
+        return data.encode()
+    else:
+        return data
+
+
+CallbackT = Callable[[bytes], bytes]
+
+
+def _wrap_with_callback(
```
jakubroztocil, line 55: Can we come up with a better name?

httpie/uploads.py
```diff
@@ -36,69 +42,101 @@ def __iter__(self) -> Iterable[Union[str, bytes]]:
...
+) -> Callable[..., Any]:
+    @functools.wraps(func)
+    def wrapped(*args, **kwargs):
+        chunk = func(*args, **kwargs)
+        callback(chunk)
+        return chunk
+    return wrapped
+
+
+def _process_file(
```
jakubroztocil, line 67: Can we come up with a better name?

PR #1256, opened 2021-12-30
Add warnings when there is no incoming data from stdin

### LINKED ISSUE #1255: Warn the user when there is no incoming data after a certain time passed on stdin
E.g
```
 $ cat | http POST pie.dev/post                                                                                                                                             861ms
> no stdin data read in 10.0s (perhaps you want to --ignore-stdin)
> https://httpie.io/docs/cli/best-practices

```

isidentical:
Fixes #1255

### DIFF
--- a/httpie/core.py
+++ b/httpie/core.py
@@ -188,7 +188,7 @@ def program(args: argparse.Namespace, env: Environment) -> ExitStatus:
             args.follow = True  # --download implies --follow.
             downloader = Downloader(output_file=args.output_file, progress_file=env.stderr, resume=args.download_resume)
             downloader.pre_request(args.headers)
-        messages = collect_messages(env, args=args, config_dir=env.config.directory,
+        messages = collect_messages(env, args=args,
                                     request_body_read_callback=request_body_read_callback)
         force_separator = False
         prev_with_body = False
--- a/httpie/core.py
+++ b/httpie/core.py
@@ -188,7 +188,7 @@ def program(args: argparse.Namespace, env: Environment) -> ExitStatus:
             args.follow = True  # --download implies --follow.
             downloader = Downloader(output_file=args.output_file, progress_file=env.stderr, resume=args.download_resume)
             downloader.pre_request(args.headers)
-        messages = collect_messages(args=args, config_dir=env.config.directory,
+        messages = collect_messages(env, args=args, config_dir=env.config.directory,
                                     request_body_read_callback=request_body_read_callback)
         force_separator = False
         prev_with_body = False
--- a/httpie/uploads.py
+++ b/httpie/uploads.py
@@ -1,3 +1,5 @@
+import sys
+import os
 import zlib
 import functools
 from typing import Any, Callable, IO, Iterable, Optional, Tuple, Union, TYPE_CHECKING
@@ -9,6 +11,7 @@ from requests.utils import super_len
 if TYPE_CHECKING:
     from requests_toolbelt import MultipartEncoder
 
+from .context import Environment
 from .cli.dicts import MultipartRequestDataDict, RequestDataDict
 
 
@@ -64,13 +67,47 @@ def _wrap_function_with_callback(
     return wrapped
 
 
+def is_stdin(file: IO) -> bool:
+    try:
+        file_no = file.fileno()
+    except Exception:
+        return False
+    else:
+        return file_no == sys.stdin.fileno()
+
+
+READ_THRESHOLD = float(os.getenv("HTTPIE_STDIN_READ_THRESHOLD", 10.0))
+
+
+def trigger_stdin_warning_thread(env: Environment, file: IO) -> None:
+    import select
+    import threading
+
+    def worker():
+        can_read, _, _ = select.select([file], [], [], READ_THRESHOLD)
+        if not can_read:
+            env.stderr.write(
+                f"> no stdin data read in {READ_THRESHOLD}s "
+                f"(perhaps you want to --ignore-stdin)\n"
+            )
+            env.stderr.write("> https://httpie.io/docs/cli/best-practices\n")
+
+    thread = threading.Thread(
+        target=worker
+    )
+    thread.start()
+
+
 def _prepare_file_for_upload(
+    env: Environment,
     file: Union[IO, 'MultipartEncoder'],
     callback: CallbackT,
     chunked: bool = False,
     content_length_header_value: Optional[int] = None,
 ) -> Union[bytes, IO, ChunkedStream]:
     if not super_len(file):
+        if is_stdin(file):
+            trigger_stdin_warning_thread(env, file)
         # Zero-length -> assume stdin.
         if content_length_header_value is None and not chunked:
             # Read the whole stdin to determine `Content-Length`.
@@ -103,6 +140,7 @@ def _prepare_file_for_upload(
 
 
 def prepare_request_body(
+    env: Environment,
     raw_body: Union[str, bytes, IO, 'MultipartEncoder', RequestDataDict],
     body_read_callback: CallbackT,
     offline: bool = False,
@@ -125,6 +163,7 @@ def prepare_request_body(
 
     if is_file_like:
         return _prepare_file_for_upload(
+            env,
             body,
             chunked=chunked,
             callback=body_read_callback,
--- a/httpie/uploads.py
+++ b/httpie/uploads.py
@@ -80,6 +80,12 @@ READ_THRESHOLD = float(os.getenv("HTTPIE_STDIN_READ_THRESHOLD", 10.0))
 
 
 def trigger_stdin_warning_thread(env: Environment, file: IO) -> None:
+    # Windows unfortunately does not support select() operation
+    # on regular files, like stdin in our use case.
+    # https://docs.python.org/3/library/select.html#select.select
+    if os.name == 'nt':
+        return None
+
     import select
     import threading
 
--- a/httpie/uploads.py
+++ b/httpie/uploads.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
 
 from .context import Environment
 from .cli.dicts import MultipartRequestDataDict, RequestDataDict
+from .compat import is_windows
 
 
 class ChunkedStream:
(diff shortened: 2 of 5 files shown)


maldito187 (commented):
https --debug -F drwg.ru HTTPie 2.6.0 Requests 2.26.0 Pygments 2.10.0 Python 3.10.0 (tags/v3.10.0:b494f59, Oct 4 2021, 19:00:18) [MSC v.1929 64 bit (AMD64)] D:\Python310\python.exe Windows 10 <Environment {'colors': 256, 'config': {'default_options': []}, 'config_dir': WindowsPath('C:/Users/kkrasnov/AppData/Roaming/httpie'), 'devnull': <property object at 0x00000232C296B150>, 'is_windows': True, 'log_error': <function Environment.log_error at 0x00000232C29B1750>, 'program_name': 'https', 'stderr': <colorama.ansitowin32.StreamWrapper object at 0x00000232C2984C70>, 'stderr_isatty': True, 'stdin': <_io.TextIOWrapper name='<stdin>' mode='r' encoding='utf-8'>, 'stdin_encoding': 'utf-8', 'stdin_isatty': True, 'stdout': <colorama.ansitowin32.StreamWrapper object at 0x00000232C2984610>, 'stdout_encoding': 'utf-8', 'stdout_isatty': True}> <PluginManager {'adapters': [], 'auth': [<class 'httpie.plugins.builtin.BasicAuthPlugin'>, <class 'httpie.plugins.builtin.DigestAuthPlugin'>], 'converters': [], 'formatters': [<class 'httpie.output.formatters.headers.HeadersFormatter'>, <class 'httpie.output.formatters.json.JSONFormatter'>, <class 'httpie.output.formatters.xml.XMLFormatter'>,
@maldito-dot

luisuimi:
Continue to review full report at Codecov.

Legend - Click here to learn more
Δ = absolute <relative> (impact), ø = not affected, ? = missing data
Powered by Codecov. Last update 4c56d89...18eabfb. Read the comment docs.

httpie/uploads.py
```diff
@@ -64,13 +67,53 @@ def wrapped(*args, **kwargs):
...
+    import threading
+
+    def worker():
+        can_read, _, _ = select.select([file], [], [], READ_THRESHOLD)
+        if not can_read:
+            env.stderr.write(
+                f"> no stdin data read in {READ_THRESHOLD}s "
+                f"(perhaps you want to --ignore-stdin)\n"
+            )
+            env.stderr.write("> https://httpie.io/docs/cli/best-practices\n")
```
jakubroztocil, line 99: We should probably use the existing `env.log_error(msg, level='warning')`
https://github.com/httpie/httpie/blob/245cede2c274bef7b172fe65814bfabc6f87b0b4/httpie/context.py#L137-L139

httpie/core.py
```diff
@@ -188,7 +188,7 @@ def request_body_read_callback(chunk: bytes):
             args.follow = True  # --download implies --follow.
             downloader = Downloader(output_file=args.output_file, progress_file=env.stderr, resume=args.download_resume)
             downloader.pre_request(args.headers)
-        messages = collect_messages(args=args, config_dir=env.config.directory,
+        messages = collect_messages(env, args=args, config_dir=env.config.directory,
```
jakubroztocil, line 191: Here and elsewhere: Could we be more specific in what the functions need from `env`? If it would be too cumbersome, then we could probably remove the specific args like `config_dir` here.

httpie/uploads.py
jakubroztocil: And the link could be prefixed with `See: `

httpie/uploads.py
```diff
@@ -64,13 +67,53 @@ def wrapped(*args, **kwargs):
...
+
+READ_THRESHOLD = float(os.getenv("HTTPIE_STDIN_READ_THRESHOLD", 10.0))
+
+
+def trigger_stdin_warning_thread(env: Environment, file: IO) -> None:
+    # Windows unfortunately does not support select() operation
+    # on regular files, like stdin in our use case.
+    # https://docs.python.org/3/library/select.html#select.select
+    if os.name == 'nt':
+        return None
```
jakubroztocil, line 87: We have `httpie.compat.is_windows()`. If it’s more nuanced, I’d still probably add a named check.

httpie/uploads.py
```diff
@@ -64,13 +67,53 @@ def wrapped(*args, **kwargs):
...
+    except Exception:
+        return False
+    else:
+        return file_no == sys.stdin.fileno()
+
+
+READ_THRESHOLD = float(os.getenv("HTTPIE_STDIN_READ_THRESHOLD", 10.0))
+
+
+def trigger_stdin_warning_thread(env: Environment, file: IO) -> None:
```
jakubroztocil, line 82: Maybe something like `observe_stdin_for_data_thread`?

jakubroztocil (changes requested):

httpie/uploads.py
```diff
@@ -64,13 +67,53 @@ def wrapped(*args, **kwargs):
...
+def is_stdin(file: IO) -> bool:
+    try:
+        file_no = file.fileno()
+    except Exception:
+        return False
+    else:
+        return file_no == sys.stdin.fileno()
+
+
+READ_THRESHOLD = float(os.getenv("HTTPIE_STDIN_READ_THRESHOLD", 10.0))
```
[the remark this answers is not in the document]
isidentical: > I wonder: Should make this configurable (--xxx=60) & disableable (--xxx=0)?

I am not sure if it would be useful at all, considering this behavior has no side-effect (e.g, we do not terminate the process), just a warning. I explicitly ignored to have this documented, since this is primarily a test-only env var (there is no way to mock stuff on tests where I need to create subprocesses). I'd prefer make it just an invisible configuration var, until someone official asks whether they can disable it or not.

> Also, single quotes (many places).

My bad :/ I'll integrate the [pre-commit-hooks/double-quote-fixer](https://github.com/pre-commit/pre-commit-hooks#double-quote-string-fixer) right after 3.0 release, so we won't need to worry about this.

httpie/core.py
isidentical: > Here and elsewhere: Could we be more specific in what the functions need from env? If it would be too cumbersome, then we could probably remove the specific args like config_dir here.

Sure thing. I did not notice `config_dir` was simply an attribute of `env`.

httpie/uploads.py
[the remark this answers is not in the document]
jakubroztocil: > I am not sure if it would be useful at all, considering this behavior has no side-effect

If you do run this from a `cron` job like `/bin/slow-script | http POST pie.dev/post`, then the warning will have a side-effect in the form of an email, for example.

Also, I believe the current implementation doesn’t respect [`-qq`](https://httpie.io/docs/cli/verbose-output).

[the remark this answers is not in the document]
isidentical: > If you do run this from a cron job like /bin/slow-script | http POST pie.dev/post, then the warning will have a side-effect in the form of an email, for example.

Am I missing something:
```console
$ python -c 'import time; time.sleep(2); print("regular data")' | HTTPIE_STDIN_READ_THRESHOLD=1 http -p=B POST pie.dev/post                                                                                                                                                        2,345s

http: warning: > no stdin data read in 1.0s (perhaps you want to --ignore-stdin)
> See: https://httpie.io/docs/cli/best-practices

regular data


```
This does not affect the request body.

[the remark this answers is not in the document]
isidentical: > Also, I believe the current implementation doesn’t respect -qq.

👍🏻 TIL. Will add support for this one.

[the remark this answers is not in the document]
jakubroztocil: I’m referring to the behavior of `cron` where any output triggers an email being sent (an example of a side-effect the output can have):

``` 
$ man cron
```

> The cron utility then wakes up every minute, examining all stored
>      crontabs, checking each command to see if it should be run in the
>      current minute.  **When executing commands, any output is mailed to
>      the owner of the crontab** (or to the user named in the MAILTO
>      environment variable in the crontab, if such exists).
>

httpie/uploads.py
isidentical: @jakubroztocil any idea why `log_error` chooses to write to the original stderr instead of the regular one (which becomes devnull on `-qq`). If I switch, then it prints the warnings even with `-qq`.

httpie/uploads.py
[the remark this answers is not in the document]
isidentical: > I’m referring to the behavior of cron where any output triggers an email being sent (an example of a side-effect the output can have — admittedly a confusing one):

Ah, that makes sense. Then the question becomes whether we should only support `-qq` or add an additional configuration option. I'd go with `-qq` (which I believe is already supported).

httpie/uploads.py
jakubroztocil: I vaguely remember there being one, likely related to  `HTTPieArgumentParser._setup_standard_streams()`:

https://github.com/httpie/httpie/blob/17ed3bb8c519fc6e0cc62fa3bb0fc10929de2595/httpie/cli/argparser.py#L197-L235

Take a look. Otherwise, it would be the natural approach 👍🏻

httpie/core.py
jakubroztocil: For the record, there’s a strong preference for being specific because, without it, it’s really hard to understand what the function tree depends on, which makes reasoning about many things harder.

httpie/uploads.py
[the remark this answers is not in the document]
jakubroztocil: Alright, let’s keep things simple and rely on `--qq` for now. We can revisit this if we receive some complaints about this warning. But we should probably handle `HTTPIE_STDIN_READ_THRESHOLD=0` despite that not being documented (also, this might be a slightly better name: `HTTPIE_STDIN_READ_WARN_THRESHOLD`).

tests/test_uploads.py
```diff
@@ -83,6 +88,69 @@ def test_chunked_raw(httpbin_with_chunked_support):
...
+    )
+    try:
+        yield process_1, process_2
+    finally:
+        process_1.terminate()
+        process_2.terminate()
+
+
+@pytest.mark.parametrize("wait", (True, False))
+@pytest.mark.skipif(os.name == 'nt', reason="Windows doesn't support select() calls into files")
```
jakubroztocil, line 122: `is_windows()`

httpie/uploads.py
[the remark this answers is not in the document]
jakubroztocil: We could also prefix the name with an underscore to _underscore_ it’s an internal variable.

httpie/uploads.py
isidentical: Seems like some tests (on the download logic) was depending on this behaviour. Currently leaving as is and printing to `stderr` directly for now (to support `-qq`).

httpie/uploads.py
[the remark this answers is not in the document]
isidentical: > We could also prefix the name with an underscore to underscore it’s an internal variable.

The Python level name or the environment variable? I don't think I've ever seen an environment variable with an underscore prefix.

maldito187 (commented):
@maldito-dot

PR #1260, opened 2022-01-04
Re-design nested JSON spec

isidentical:


jakubroztocil (changes requested):
Great stuff 🙌🏻  I’d just introduce some constants for things like `[`, `]`, `\\`, `append`, etc. 

I’m polishing the docs in the meantime, and might tweak the error message a bit — will push the changes to this PR.

isidentical:
Sure, waiting for you. As you have probably noticed, for the constants `[`/`]` etc. we already use tokens whenever we can, so their usage is pretty limited. And I think `path.kind` can be an enum, but not a really high priority.

jakubroztocil:
> Sure, waiting for you. As you have probably noticed, for the constants `[`/`]` etc. we already use tokens whenever we can, so their usage is pretty limited. And I think `path.kind` can be an enum, but not a really high priority.

My point is that there’s a lot of unnamed string references to strings with a special meaning. The code will be easier to understand and navigate if we consistently use named constants/enums.

isidentical:
> My point is that there’s a lot of unnamed string references to strings with a special meaning. The code will be easier to understand and navigate if we consistently use named constants/enums.

I'll try to enumerize the `kind`s (`append`/`index`/`key`/`set`).

PR #1277, opened 2022-01-23
Fix time elapsed

jakubroztocil:
- [x] show the actual time elapsed instead of time-to-parse-headers
- [x] add docs for response meta

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -538,7 +538,7 @@ and URL parameters. This is a very practical way of constructing
 HTTP requests from scratch on the CLI.
 
 Each *request item* is simply a key/value pair separated with the following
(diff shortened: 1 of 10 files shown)


isidentical (approved):
You are faster than me 😄 (I was planning this on monday). The conditional body download seems a bit weird, considering how output options can change the total elapsed time but I guess there is no easy solution on that front. Maybe we might consider adding a warning in the future (`Elapsed Time: 1.2042824 seconds (excluding time spent on the response body)`) or using different keys but since it is documented I guess it is not a big deal.

jakubroztocil:
Yeah, the conditional body download, which is otherwise a good thing, makes the metrics less robust. The notice might be a good idea for a future iteration.

PR #1303, opened 2022-02-25
Don't issue redundant stdin detection warning when  is in place.

isidentical:
This patch fixes a race happening between 2 IO bound operations:
- `read(2)` with `sys.stdin`
- `select(2)` with `sys.stdin`

image

`select(2)` is a function that takes files and tells when they are ready to be processed (read/written etc.). It is used for concurrently dealing with multiple sockets (e.g you have multiple open connections, and with `select(2)` you can check which one of those connections have some data to be processed). In HTTPie however, we used it to detect whether there is any incoming data to the `sys.stdin`, so that when there is no data we could issue a warning.

The problem is that, the initial version of this code unknowlingly created a race between the `select.select` call:
https://github.com/httpie/httpie/blob/30cd862fc0e173698fc17487c4b96d8f64b701ea/httpie/uploads.py#L98-L110

and the `file.read()` call:
https://github.com/httpie/httpie/blob/30cd862fc0e173698fc17487c4b96d8f64b701ea/httpie/uploads.py#L124-L132

Both of these functions are blocking, and we ran them concurrently to get away from that issue. But since Python has a GIL, when you run multiple threads together, depending on the CPU-related activity and some other details it switches between them on certain intervals. (green arrows indicate the interpreter given the execution to the Main Thread, red indicate it is given to Observer Thread)

image

So to put it simply, we used to run the `read()` and the `select()` at the same time and whichever gets choosen to be executed at that time would win. If it was `select()`, we'd get the correct result (about whether there is any data or not) and if it was `read()` we'd always get the wrong result (no incomiong data). This caused to intermittent / random failure. From what I understood this problem happens randomly (and in a OS-dependant way, because of the underlying thread scheduler), and there is no fair way of testing it (except manually). 

The race has been solved by moving the `select(2)` out of the thread (as a guard) and using `threading.Event`s as indicators about whether we've seen any data or not.

### DIFF
--- a/httpie/uploads.py
+++ b/httpie/uploads.py
@@ -2,6 +2,8 @@ import sys
 import os
 import zlib
 import functools
+import time
+import threading
 from typing import Any, Callable, IO, Iterable, Optional, Tuple, Union, TYPE_CHECKING
 from urllib.parse import urlencode
 
@@ -22,12 +24,20 @@ class ChunkedStream:
 
 
 class ChunkedUploadStream(ChunkedStream):
-    def __init__(self, stream: Iterable, callback: Callable):
+    def __init__(
+        self,
+        stream: Iterable,
+        callback: Callable,
+        event: Optional[threading.Event] = None
+    ) -> None:
         self.callback = callback
         self.stream = stream
+        self.event = event
 
     def __iter__(self) -> Iterable[Union[str, bytes]]:
         for chunk in self.stream:
+            if self.event:
+                self.event.set()
             self.callback(chunk)
             yield chunk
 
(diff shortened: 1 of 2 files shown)


httpie/uploads.py
```diff
@@ -80,7 +97,7 @@ def is_stdin(file: IO) -> bool:
 READ_THRESHOLD = float(os.getenv("HTTPIE_STDIN_READ_WARN_THRESHOLD", 10.0))
```
jakubroztocil, line 97:
```suggestion
READ_THRESHOLD = float(os.getenv('HTTPIE_STDIN_READ_WARN_THRESHOLD', 10.0))
```

jakubroztocil (commented):
Good stuff. Let’s just add a mention to CHANGELOG

PR #1304, opened 2022-02-25
Fix bug #662 - cookies not being maintained correctly over session involving redirects+set-cookie - need to save cookie domain in sessions file

barny:
Observed problems using --follow with --session, that on resuming using the saved session but expired authentication cookies stored in the session, httpie was generating duplicated cookies - saving the domain in the session file resolved this.


isidentical:
Thanks for the PR @barny, but we have decided to take a different approach to fully solve this problem in #1312.

Issue #1305, opened 2022-02-28
Documentation suggests running potentially harmful operation on Arch-based distros


stasiaks:
Installation documentation for [Arch Linux](https://httpie.io/docs/cli/arch-linux) tells you to run

> pacman -Sy httpie

in order to install httpie. I understand that it was possibly done so that newbie doesn't end up with bunch of 404s thanks to new versions being released since their last upgrade. But doing so will perform a **partial upgrade**, which is not only unsupported but heavily discouraged (especially since httpie does have dependencies that are very popular, like `python-setuptools` or `python-requests`)

ArchWiki states [here](https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported) that:

> (...) if two packages depend on the same library, upgrading only one package might also upgrade the library (as a dependency), which might then break the other package which depends on an older version of the library.
>
>That is why partial upgrades are **not supported**. Do not use:
>
> - **pacman -Sy package**
> - pacman -Sy followed by pacman -S package.
> - pacman -Syuw (Note that pacman -Syuw does imply the same risks like pacman -Sy, as it will update the pacman sync database without installing the newer packages.)

Sure, experienced user will know not to do that, but experienced user didn't need that command in first place, while newbie might end up with their dependencies broken, and no clue on how to fix them.

Thus I'd suggest changing [that section of documentation](https://httpie.io/docs/cli/arch-linux) to something along these lines:

```md

Also works for other Arch-derived distributions like ArcoLinux, EndeavourOS, Artix Linux, etc.

# Install httpie
pacman -S httpie

# Upgrade httpie by performing full upgrade
pacman -Syu 
```

isidentical:
Hey @stasiaks, thanks for opening up this issue! I agree in your judgement (and thanks for sharing the official documentation to corroborate that). Would you like to submit a patch to fix this? A simple fix would be

- Updating the arch section in [`methods.yaml`](https://github.com/httpie/httpie/blob/master/docs/installation/methods.yml)
- Running `python generate.py` in `docs/installation` to update the documentation.

stasiaks:
Hi, I'd be glad to, I've created PR for this (#1306)

PR #1306, opened 2022-02-28
Change partial upgrade to full upgrade in Arch Linux installation documentation

stasiaks:
Addresses problem described in #1305 changing command from partial upgrade to full upgrade on installation, which should be both safe and painless for new into rolling-release model (`-S` alone could cause a stream of **404** responses due to old versions not being stored)

Additionally removed argument from upgrade command, since `-Syu` alone will update package that's already installed.


isidentical:
Thanks a lot for the PR @stasiaks!

isidentical (approved):

PR #1309, opened 2022-03-01
Implement support for private key passphrases

### LINKED ISSUE #946: Support for client side certificate/key password
Hi,

I don't know if it is on the roadmap, but support for certificate/key password options would be awesome.

Right now httpie triggers std input to get a password, but it's not ideal in scripting.

isidentical:
Resolves #946.


httpie/cli/definition.py
```diff
@@ -803,6 +803,17 @@ def format_auth_help(auth_plugins_mapping):
     '''
 )
 
+ssl.add_argument(
+    '--cert-key-pass',
+    default=None,
+    type=SSLCredentials,
+    help='''
+    The passphrase to be used to with the given private key. Only needed if --cert-key
+    is given and the key file requires a passphrase.
```
jakubroztocil, line 812: > If not provided, you’ll be prompted interactively

PR #1314, opened 2022-03-08
Add table headers for upgrade flags

isidentical:


isidentical:
@jakubroztocil seems like GitHub does not allow tables without headers: https://stackoverflow.com/questions/17536216/create-a-table-without-a-header-in-markdown

PR #1327, opened 2022-03-15
Add support for sending secure cookies over localhost

### LINKED ISSUE #1308: Invaliding Secure Cookie handling with localhost
## Checklist

- [X] I've searched for similar issues.
- [X] I'm using the latest version of HTTPie.

---

Given the following tiny Go HTTP server
```GO
package main

import (
	"net/http"
)

func set(rw http.ResponseWriter, _ *http.Request) {
	http.SetCookie(rw, &http.Cookie{
		Name:     "session-token",
		Value:    "token here",
		Secure:   true,
		HttpOnly: false,
		SameSite: http.SameSiteNoneMode,
	})
	rw.Write(nil)
}

func get(rw http.ResponseWriter, req *http.Request) {
	cookie, err := req.Cookie("session-token")
	if err != nil {
		rw.Write(nil)
		return
	}
	rw.Write([]byte(cookie.Value))
}

func main() {
	http.HandleFunc("/set-cookie", set)
	http.HandleFunc("/get-cookie", get)

	panic(http.ListenAndServe(":8090", nil))
}
```
If you set the cookie to a session with the following command
```bash
http --session=./session.json GET http://localhost:8090/set-cookie
```
I would expect to be able to get the cookie value by the next running 
```bash
http --session=./session.json GET http://localhost:8090/get-cookie
```

However, the cookie is unset in the second request if the Secure flag is set to true.

As per the spec; 
A cookie with the Secure attribute is only sent to the server with an encrypted request over the HTTPS protocol. It's never sent with unsecured HTTP (***except on localhost***), which means [man-in-the-middle](https://developer.mozilla.org/en-US/docs/Glossary/MitM) attackers can't access it easily. (https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies)

This appears to be mishandled in httpie, as it is not sending secure cookies when localhost is used.

# Expected Behavior
Secure cookies are sent in localhost requests even if https is not in use

# Actual Behavior
Secure cookies are not set

isidentical:
Fixes #1308. See the discussions in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1618113) and [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1056543) for more context.

P.S: We do not support localhost aliases. As far as I understand, Firefox also uses a similiar technique for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added support for session persistence of repeated headers with the same name. ([#1335](https://github.com/httpie/httpie/pull/1335))
 - Changed `httpie plugins` to the new `httpie cli` namespace as `httpie cli plugins` (`httpie plugins` continues to work as a hidden alias). ([#1320](https://github.com/httpie/httpie/issues/1320))
 - Fixed redundant creation of `Content-Length` header on `OPTIONS` requests. ([#1310](https://github.com/httpie/httpie/issues/1310))
-
+- Added support for sending `Secure` cookies to the `localhost` (and `.local` suffixed domains). ([#1308](https://github.com/httpie/httpie/issues/1308))
 
 ## [3.1.0](https://github.com/httpie/httpie/compare/3.0.2...3.1.0) (2022-03-08)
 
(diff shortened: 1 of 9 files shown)


jakubroztocil (changes requested):
There are now some failing tests and conflicts.

PR #1336, opened 2022-03-29
Automatic release update warnings.

isidentical:
This PR embeds an automatic **update warning system** into the HTTPie. What this means is that, on some configurable thresholds (once a week) HTTPie will check whether a new release is available and if it there is it will give user a warning.

Here is an example:
image

This problem consists of two dimensions, which each need to be handled separately:
1. Correctly (without any false positives) detect whether there is a new release is available.
2. Ensure there is no observable startup penalty.

### Update detection

Since HTTPie is distributed on many different channels (from Homebrew to Snap, PyPI to Chocolatey) we need to be careful to show these warnings to only users whom can upgrade at that moment. So for that, we have a new file called `httpie/internal/_build_channel.py`. This file has no runtime purpose, other than contain the `BUILD_CHANNEL` constant.

The default for that constant is `unknown`, and each build system will overwrite it with the value pointing to their name. E.g on Homebrew, during the build process, we'll write `BUILD_CHANNEL = 'brew'` into that file. This way, we can detect whether the build is from Homebrew or not.

Once we know what build channel the user is using, we need to determine whether we have a new release for that channel. That is where [`packages.httpie.io/latest.json`](packages.httpie.io/latest.json) comes into place. It is an automatically updated (in a daily basis) registry of build channels and the HTTPie versions in them.

We compare the current version of HTTPie with the latest version for the build channel by looking to the `latest.json`. If there is a new release, we show the warning.

### Startup penalty

For reducing the overhead of this process as much as we can, we do it in two steps.

- Fetch the `latest.json` into a local file (`~/.config/httpie/version_info.json`) (done in a separate process, takes a long time)
- Warn the user (done in the same process, takes literally no time <1ms)

For the first task, this PR introduces a daemon system where we can start tasks in the background and detach them from the main process. So even if you run a short HTTPie command (e.g `http --version`), you exit immediately and the actual task still works to finish up its objective without blocking you.

This part is implemented by using [double-forking](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap11.html#tag_11_01_03) on POSIX, and creating a subprocess on Windows.

#### Baseline

Master:
```
BASELINE (master)# hyperfine 'http --offline pie.dev/get' 
Benchmark 1: http --offline pie.dev/get
  Time (mean ± σ):     217.7 ms ±   4.8 ms    [User: 199.0 ms, System: 18.1 ms]
  Range (min … max):   210.1 ms … 225.0 ms    13 runs
```

This branch:
```
(this branch)# hyperfine 'http --offline pie.dev/get'
Benchmark 1: http --offline pie.dev/get
  Time (mean ± σ):     217.0 ms ±   5.4 ms    [User: 195.7 ms, System: 21.0 ms]
  Range (min … max):   210.3 ms … 226.9 ms    13 runs

(this branch)# hyperfine 'rm ~/.config/httpie/version_info.json; http --offline pie.dev/get'
Benchmark 1: rm ~/.config/httpie/version_info.json; http --offline pie.dev/get
  Time (mean ± σ):     236.8 ms ±  14.5 ms    [User: 205.3 ms, System: 30.9 ms]
  Range (min … max):   213.7 ms … 253.3 ms    11 runs
```

It seems like once the `version_info` file is created, there is no penalty. When the file does not exist, there is a very small cost of forking the process which is ~15 ms (you will pay this cost twice a month, so if you ran HTTPie 1000 times that month this will cost ~30 ms in overall overhead).


isidentical:
This is blocked by #1330

iisamariisamii (approved):

PR #1345, opened 2022-04-01
Ping werkzeug to <2.1.0

isidentical:
The new release breaks `pytest-httpbin`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ tests_require = [
     'pytest-httpbin>=0.0.6',
     'pytest-lazy-fixture>=0.0.6',
     'responses',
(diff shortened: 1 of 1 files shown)


davidism:
Would you open an issue to track updating compatibility?

isidentical:
Hey @davidism! I think our failure originates from https://github.com/postmanlabs/httpbin/issues/673. We use https://github.com/kevin1024/pytest-httpbin on our tests, which directly depends to https://github.com/postmanlabs/httpbin.

davidism:
It looks like httpbin is no longer maintained, are there any maintained forks? I guess this is only for test requirements, but it's still not ideal long term.

PR #1356, opened 2022-04-15
Docs/ Installation with Pacman (Arch Linux)

leomp12:
Hey guys,
Current documentation is instructing to upgrade all packages with `pacman -Syu` to install/upgrade httpie, it's not necessary at all and I think it's most appropriate docs instructing to install/upgrade only httpie package, right?
Also, Pacman install must be run with sudo because it requires root and does not ask for authentication after called.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -204,13 +204,8 @@ $ yum upgrade httpie
 Also works for other Arch-derived distributions like ArcoLinux, EndeavourOS, Artix Linux, etc.
 
 ```bash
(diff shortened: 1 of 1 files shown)


isidentical:
We changed it from that, since partial upgrades are not recommended. Please see #1306 (by the way if the reasoning is not sufficient, please let me know. I am not an arch user myself, but the problem described in #1305 seems important enough for us to *not to* suggest a partial upgrade).

Issue #1362, opened 2022-04-18
star history~


kalsolio:
If you have a general question, please consider asking on Discord: https://httpie.io/chat
## Star History

[![Star History Chart](https://api.star-history.com/svg?repos=httpie/httpie&type=Date)](https://star-history.com/#httpie/httpie&Date)

isidentical:
Thanks for posting this, more details here https://httpie.io/blog/stardust

PR #1364, opened 2022-04-18
Man page fixes

isidentical:
- Highlighting for options (`-x`, `--x`) now doesn't strip the prefix (may be whitespace).
- Escape sequences are now cross-platform compatible (directly taken by groff/troff [man's renderer])
- Now we check for the section before displaying the man pages. On MacOS, there is `HTTP(n)` which is different from our `HTTP(1)`. This used to conflict with it, and we showed the wrong page. Now we specifically ask foir `HTTP(1)`.
- Errors that might happen (e.g non executable man command) is now suppressed. So in the worst case (if anything regarding man execution goes wrong), we'll always display the manual.
- Docs for man pages.
- Man page for `httpie` (and it's own custom part telling people to use `man http`)
- Epilog for the man pages (see also part)
- Comments warning you should not edit the file, and pointing to the right source.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -277,7 +277,7 @@ Synopsis:
 $ http [flags] [METHOD] URL [ITEM [ITEM]]
 ```
 
-See also `http --help`.
+See also `http --help` (and for systems where man pages are available, you can use `man http`).
 
 ### Examples
 
--- a/extras/man/http.1
+++ b/extras/man/http.1
@@ -1,3 +1,4 @@
+.\" This file is auto-generated from the parser declaration in httpie/cli/definition.py by extras/scripts/generate_man_pages.py.
 .TH http 1 "2022-03-08" "HTTPie 3.1.1.dev0" "HTTPie Manual"
 .SH NAME
 http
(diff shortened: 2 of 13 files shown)


repdox-cyfso6-wonzAb:
Duplicate  #1362

repdox-cyfso6-wonzAb:
Duplicate of #

petersadek90 (approved):

petersadek90 (approved):

jakubroztocil (commented):
We talked also about `man httpie`.

isidentical:
We should have that, no?

jakubroztocil:
Yes, even if just something super simple. E.g., a list of the management commands and a clear mention of `man http` / `man https`. It’s important to have it because people will tend to run `man httpie`. (Sorry if it’s already there, I’m yet to carefuly review this.)

isidentical:
Ah, ok. That is what I meant, it is already here.

isidentical:
We were directly using the description which mentioned `http` and `https` commands, but not `man http` explicitly. I've also added it now. 
image

jakubroztocil:
@isidentical that’s perfect 🤩

petersadek90 (approved):

PR #1365, opened 2022-04-19
Remove unnecessary $ in bash example

MrBrain295:


petersadek90 (approved):

petersadek90 (approved):

isidentical:
The `$` denotes it is a shell command. It is the general practice for denoting it, so I am not really sure why we should remove it. Closing it for now, but if there is something problematic please let us know.

MrBrain295:
It just makes it easier when copying commands.

jakubroztocil:
Thanks, @MrBrain295. You’re right that it makes the copying slightly harder, but it’s also a visual cue that it’s indeed a shell command as @isidentical points out. Btw, in our docs, which we build from `./docs/README.md`, we make the `$` un-selectable and exclude it when copying the command.

PR #1375, opened 2022-04-28
Disable PackIt CI on the PRs

isidentical:
We still use other functionalities (such as automatically warning downstream people when a release is ready) but the PR checks are randomly failing and the errors are very hard to decipher.


hroncok:
:(

Can I do something about this? Do you have example failures I can look at?

hroncok:
FTR I get notified on new releases even without packit, so keeping it just for that might be redundant.

PR #1377, opened 2022-04-28
wip implementation of --base-url argument

vergenzt:
Addresses httpie/httpie#838

Doesn't have tests yet, nor has anyone else weighed in on approach, naming, etc. But it's simple and is working well for me, so I figured I'd share in case others want to help out with the remaining boring bits! 😄 

I'm using locally by uninstalling my default (Homebrew) httpie installation, and re-installing this version via:

```
$ pipx install git+https://github.com/vergenzt/httpie.git@issue-838-add-base-url
```

### DIFF
--- a/httpie/cli/argparser.py
+++ b/httpie/cli/argparser.py
@@ -414,38 +414,10 @@ class HTTPieArgumentParser(BaseHTTPieArgumentParser):
 
         """
         if self.args.method is None:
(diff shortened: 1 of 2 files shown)


zachvalenta:
Having a `--base-url` option would be great.

Herry0092 (approved):

Herry0092 (approved):

PR #1383, opened 2022-05-03
Checking headers to determine auto-streaming

egleston:
The header-check for auto-streaming should parse out the type/sub-type from optional parameters.
FastAPI/Starlette modifies the Content-Type header for SSE responses (actually, for all `text/*` media) by adding the `charset` as an "optional parameter": `Content-Type: text/event-stream; charset=utf-8`

### DIFF
--- a/httpie/output/writer.py
+++ b/httpie/output/writer.py
@@ -164,8 +164,10 @@ def get_stream_type_and_kwargs(
     if not is_stream and message_type is HTTPResponse:
         # If this is a response, then check the headers for determining
         # auto-streaming.
-        ct, _ = parse_content_type_header(headers.get('Content-Type'))
-        is_stream = ct == 'text/event-stream'
+        ct_raw = headers.get('Content-Type', None)
+        if ct_raw:
+            ct, _ = parse_content_type_header(ct_raw)
+            is_stream = ct == 'text/event-stream'
 
     if not env.stdout_isatty and not prettify_groups:
         stream_class = RawStream
--- a/httpie/output/writer.py
+++ b/httpie/output/writer.py
@@ -17,6 +17,7 @@ from .processing import Conversion, Formatting
 from .streams import (
     BaseStream, BufferedPrettyStream, EncodedStream, PrettyStream, RawStream,
 )
+from ..utils import parse_content_type_header
 
 
 MESSAGE_SEPARATOR = '\n\n'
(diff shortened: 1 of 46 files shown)


httpie/output/writer.py
```diff
@@ -163,7 +164,10 @@ def get_stream_type_and_kwargs(
     if not is_stream and message_type is HTTPResponse:
         # If this is a response, then check the headers for determining
         # auto-streaming.
-        is_stream = headers.get('Content-Type') == 'text/event-stream'
+        ct_raw = headers.get('Content-Type', None)
+        if ct_raw:
+            ct, _ = parse_content_type_header(ct_raw)
+            is_stream = ct == 'text/event-stream'
```
isidentical, line 170: I find expliciter names much more suitable (with the rest of the code base).
```suggestion
        raw_content_type_header = headers.get('Content-Type', None)
        if raw_content_type_header:
            content_type_header, _ = parse_content_type_header(raw_content_type_header)
            is_stream = (content_type_header == 'text/event-stream')
 ```

File: httpie/output/writer.py [hunk 2]
```diff
@@ -163,7 +164,10 @@ def get_stream_type_and_kwargs(
     if not is_stream and message_type is HTTPResponse:
         # If this is a response, then check the headers for determining
         # auto-streaming.
-        is_stream = headers.get('Content-Type') == 'text/event-stream'
+        ct_raw = headers.get('Content-Type', None)
```
isidentical, line 167: Since we reference this (and `text/event-stream`) string in different places now, it might make sense to hoist them to the global scope as a constant (just to the top of the file). E.g.:

```py
CONTENT_TYPE_HEADER_NAME = 'Content-Type'
EVENT_STREAM_TYPE = 'text/event-stream'
```

isidentical:
Overall, it looks amazing. Thanks @egleston for the PR (and finding the bug). Beside those 2 minor comments, I wouldn't touch to the HTTP server code but just rather change the test case (which should be easily adaptable):
```diff
diff --git a/tests/test_stream.py b/tests/test_stream.py
index a8950d2..8df2af2 100644
--- a/tests/test_stream.py
+++ b/tests/test_stream.py
@@ -124,6 +124,10 @@ def test_redirected_stream(httpbin):
         ['Accept:text/event-stream'],
         3
     ),
+    (
+        ['Accept:text/event-stream; charset=utf-8'],
+        3
+    ),
     (
         ['Accept:text/plain'],
         1
@@ -132,7 +136,7 @@ def test_redirected_stream(httpbin):
 def test_auto_streaming(http_server, extras, expected):
     env = MockEnvironment()
     env.stdout.write = Mock()
     http(http_server + '/drip', *extras, env=env)
     assert len([
         call_arg
         for call_arg in env.stdout.write.call_args_list
```

and add a changelog entry!

isidentical (approved):
LGTM

PR #1394, opened 2022-05-08
Updated fish completions for httpie 3.2.1

exploide:
Just updated the fish shell completions for httpie 3.2.1 and improved the existing ones a bit.


isidentical (approved):
Thanks a lot for the improvements! For the most, we actually plan to automatically generate them (e.g `--pretty` values, `--style` values, or the new options like `--manual`). But for now, I think this changes look good.

its0x08 (approved):

PR #1396, opened 2022-05-09
Fix-up standalone binary docs.

isidentical:


docs/README.md
```diff
@@ -215,14 +215,14 @@ $ pacman -Syu httpie
 $ pacman -Syu
 ```
 
-#### Single Binary Executables
+#### Single binary executables
 
-Have a standalone HTTPie executable, when you don't want to go through the full installatin process.
+Have a standalone HTTPie executable when you don't want to go through the full installation process
```
janbrasna, line 220: @isidentical I'm still trying to wrap my head around the phrasing here, is there something that escapes me from the structure of the methods above it, or the call-to-action should simply be anything like:
`"Get the standalone HTTPie Linux executables when…"` or
`"Use the standalone HTTPie Linux executables when…"`?

isidentical: That sounds much better, would you mind creating a PR for it @janbrasna?

janbrasna: @isidentical 👍 On it. Decided in favor of plurals (as there are httpie, http & https "executables" in my perception of things; not sure if there's an internal nomenclature or style guide for that), and for "Get" as a call to action. Arriving shortly.

janbrasna: refs #1399

PR #1414, opened 2022-06-17
Update Requests documentation links

nateprewitt:
This is a quick follow up from psf/requests#6140. This PR fixes the dead link by moving it to https://requests.readthedocs.io/en/latest/ which will be the primary domain for the project going forward.


jakubroztocil:
Thank you, @nateprewitt!

PR #1429, opened 2022-08-24
make code more easily to copy (fix #1417)

### LINKED ISSUE #1417: some issues with the copy button
when i click on the copy button . this what I get copied 
"# Install httpie
choco install httpie"
so I  think i will be be helpful if i can just copy the "choco install httpie"

luohuayouyi12138:
- remove `$` before bash code
- if commands are multiple, add `;`
- move the comments of out the box

english bad, see changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,25 +73,25 @@ Please note we recently accidentally made this repo private for a moment, and Gi
 Hello World:
 
 ```bash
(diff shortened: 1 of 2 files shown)


luohuayouyi12138:
sorry, edit /docs/installation/methods.yml

PR #1430, opened 2022-08-24
easier copy (#1417)

luohuayouyi12138:
the command is ran in CLI, so no `$`


chrysle:
This seems to be a duplicate with #1435

luohuayouyi12138:
> This seems to be a duplicate with #1435

the resolution is different. see my change to docs/installation/installation.jinja2. If the developers generate the document again, $ will appear again unless the change is merged.

PR #1464, opened 2023-01-02
Fix ci status badge error

shuuji3:
Hi, this PR fixes the CI badge status error due to the recent update of shields.io. You can read the details here: https://github.com/badges/shields/issues/8671.

Also, I changed to show the status of `tests` instead of `Build`.  I couldn't find `Build` workflow so I dug into the commit history and found out that the `Build` workflow was actually removed a while ago (https://github.com/httpie/httpie/commit/4c8633c6e51f388523ab4fa649040934402a4fc9#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957f). The previous badge might have been showing a false positive status. 😅 After this fix, it will show the correct status again!

**Before**
Screenshot 2023-01-03 at 1 44 42

**After**
Screenshot 2023-01-03 at 2 00 34


jakubroztocil:
@shuuji3 thank you! 🥧

shuuji3:
Thank you for merging! 🙂

PR #1468, opened 2023-01-10
Improve documentation.

amitmerchant1990:
Made the install instructions more apparent since we're not installing clients but installing HTTPie "via" several clients.


amitmerchant1990:
My bad. The copy was actually correct. Closing the PR now.

PR #1496, opened 2023-04-11
Document restriction on top-level JSON types explicitly

verhovsky:


docs/README.md
```diff
@@ -912,6 +912,8 @@ http --offline --print=B pie.dev/post \
 ]
 ```
 
+Sending other JSON types (`null`, `true`/`false`, numbers or strings) as the top level object is not possible.
```
jakubroztocil, line 915:
```suggestion
Sending scalar JSON types (a single `null`, `true`, `false`,  string or number) as the top-level object is impossible using the key/value syntax. But you can still pass it via [`--raw='<value>'`](#raw-request-body).
```

verhovsky:
@jakubroztocil you're welcome, thanks for the great docs, I was able to implement https://curlconverter.com/httpie/ quite easily.

PR #1502, opened 2023-05-07
Avoid override of headers by urllib3 when unset

abdelq:
I attempted to unset `User-Agent` when using `httpie` and couldn't understand what was going wrong as `-v` was showing no `User-Agent` until I looked at the packets and noticed that `urllib3` overwrote it.

Until something like https://github.com/psf/requests/pull/5693 makes it in `requests`, this deals with the aforementioned issue.

edit: seems like tests are failing because of the unrelated `DEFAULT_CIPHERS` issue from newer `urllib3` versions...

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -10,6 +10,7 @@ from urllib.parse import urlparse, urlunparse
 import requests
 # noinspection PyPackageRequirements
 import urllib3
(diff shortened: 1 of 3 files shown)


jakubroztocil:
Very nice! Thanks, @abdelq

PR #1505, opened 2023-05-18
fix(urllib3): :bug: could not find urllib3 DEFAULT_CIPHERS

cadesalaberry:
Default ciphers imported from urllib3 as a work around for https://github.com/httpie/httpie/issues/1499
Removed from urllib3 in this commit: https://github.com/urllib3/urllib3/commit/e5eac0c


jakubroztocil:
Thanks @cadesalaberry! Fix release coming soon.

Issue #1523, opened 2023-07-30
[macOS] Lag when piping stdout 


gsakkis:
When redirecting httpie output to a pipe, there is a significant lag in the total execution time:  

1. Without pipe
```
time http http://localhost:8508/api/v1 > /dev/null
http http://localhost:8508/api/v1> /dev/null  0.12s user 0.05s system 59% cpu 0.282 total
```

2. With pipe
```
time http http://localhost:8508/api/v1 | cat > /dev/null
http http://localhost:8508/api/v1  0.12s user 0.04s system 86% cpu 0.193 total
cat > /dev/null  0.00s user 0.00s system 0% cpu 0.973 total
```

Note that the lag appears to be happening in the `cat` command after `httpie`, which doesn't make sense and doesn't happen for `curl` or `wget`:
```
time curl -s http://localhost:8508/api/v1  > /dev/null
curl -s http://localhost:8508/api/v1 > /dev/null  0.01s user 0.01s system 56% cpu 0.024 total

time curl -s http://localhost:8508/api/v1 | cat > /dev/null
curl -s http://localhost:8508/api/v1  0.00s user 0.01s system 44% cpu 0.035 total
cat > /dev/null  0.00s user 0.00s system 9% cpu 0.034 total
```

## Checklist

- [x] I've searched for similar issues.
- [x] I'm using the latest version of HTTPie.

cedel1:
I have tried to reproduce your result but so far without meaningful success. Nothing on the scale presented by your result. My results seem to vary by server response time.

```
⬢[lukas@toolbox ~]$ time http https://restcountries.com/v3.1/alpha/cz?fields=currencies,capital > /dev/null

real	0m0.617s
user	0m0.259s
sys	0m0.050s
⬢[lukas@toolbox ~]$ time http https://restcountries.com/v3.1/alpha/cz?fields=currencies,capital | cat > /dev/null

real	0m0.641s
user	0m0.279s
sys	0m0.056s
```

```
⬢[lukas@toolbox ~]$ time http https://restcountries.com/v3.1/alpha/cz > /dev/null

real	0m0.683s
user	0m0.279s
sys	0m0.043s
⬢[lukas@toolbox ~]$ time http https://restcountries.com/v3.1/alpha/cz | cat > /dev/null

real	0m0.626s
user	0m0.271s
sys	0m0.047s

```

[...]

PR #1526, opened 2023-08-31
Remove unnecessary process wait on MacOS (issue #1523)

cedel1:
Remove `process.communicate()` in MacOS causing wait for subprocess even when the subprocess results were not used.

### DIFF
--- a/httpie/internal/daemons.py
+++ b/httpie/internal/daemons.py
@@ -87,10 +87,7 @@ def _spawn_posix(args: List[str], process_context: ProcessContext) -> None:
     if platform.system() == 'Darwin':
         # Double-fork is not reliable on MacOS, so we'll use a subprocess
         # to ensure the task is isolated properly.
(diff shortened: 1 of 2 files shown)


httpie/internal/daemons.py
```diff
@@ -87,10 +87,7 @@ def _spawn_posix(args: List[str], process_context: ProcessContext) -> None:
     if platform.system() == 'Darwin':
         # Double-fork is not reliable on MacOS, so we'll use a subprocess
         # to ensure the task is isolated properly.
-        process = _start_process(args, env=process_context)
-        # Unlike windows, since we already completed the fork procedure
-        # we can simply join the process and wait for it.
-        process.communicate()
```
cedel1, line 93: Waited for the process and could send some data to it or return its return code - but nothing of the sort is being done, basically blocking until the process finishes. But since the purpose of the process is to be daemonized (or possibly daemonized already), it doesn't seem to make sense.

PR #1540, opened 2023-11-19
Ensure support for Python 3.11/3.12

Jamim:
Hello @jkbrzt,

Thank you for this awesome tool! 🙇🏻‍♂️

I believe it would be nice to have supported Python versions listed explicitly, so I'd like to suggest changes that:
  * enable testing for Python 3.11/3.12 on CI
  * update `classifiers` at `setup.py`

Best regards!


jkbrzt:
Thanks, @Jamim!

PR #1543, opened 2023-11-26
More clear and readable expression

eukub:
Improvement of the regular expression for more clear and readable code


jkbrzt:
Thanks @eukub, but the two expressions are not equivalent.

PR #1545, opened 2023-12-09
Update README.md

geomat0101:
existing advice is broken, results in 'bash: /etc/apt/sources.list.d/httpie.list: Permission denied' because it is trying to redirect the output of the sudo command outside privileged context.  

Converted it to a subshell invocation


mikelolasagasti:
There's an alternative way to achieve the same result that I find more concise and readable. Instead of using the `sudo sh -c` approach, you could consider the following:

~~~
echo "deb [arch=amd64 signed-by=/usr/share/keyrings/httpie.gpg] https://packages.httpie.io/deb ./" |sudo tee -a /etc/apt/sources.list.d/httpie.list
~~~

jkbrzt:
Fixed via #1557. Thanks!

Issue #1547, opened 2023-12-21
Trim trailing empty lines from CLI output


olatoft:
## Checklist

- [x] I've searched for similar feature requests.

---

## Enhancement request

The output from the CLI seems to always include 2 trailing empty lines. Could they be removed?

Example command: `$ https google.com`

Current output:
```
HTTP/1.1 301 Moved Permanently
Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
Cache-Control: private, max-age=2592000
Content-Length: 220
Content-Security-Policy-Report-Only: object-src 'none';base-uri 'self';script-src 'nonce-qWBSXxuQK80KB3NcDHBMJw' 'strict-dynamic' 'report-sample' 'unsafe-eval' 'unsafe-inline' https: http:;report-uri https://csp.withgoogle.com/csp/gws/other-hp
Content-Type: text/html; charset=UTF-8
Date: Thu, 21 Dec 2023 11:01:33 GMT
Expires: Thu, 21 Dec 2023 11:01:33 GMT
Location: https://www.google.com/
P3P: CP="This is not a P3P policy! See g.co/p3phelp for more info."
Server: gws
Set-Cookie: CONSENT=PENDING+017; expires=Sat, 20-Dec-2025 11:01:33 GMT; path=/; domain=.google.com; Secure
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 0

<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
</BODY></HTML>


```

The output I would like:
```
HTTP/1.1 301 Moved Permanently
Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
Cache-Control: private, max-age=2592000
Content-Length: 220
Content-Security-Policy-Report-Only: object-src 'none';base-uri 'self';script-src 'nonce-qWBSXxuQK80KB3NcDHBMJw' 'strict-dynamic' 'report-sample' 'unsafe-eval' 'unsafe-inline' https: http:;report-uri https://csp.withgoogle.com/csp/gws/other-hp
Content-Type: text/html; charset=UTF-8
Date: Thu, 21 Dec 2023 11:01:33 GMT
Expires: Thu, 21 Dec 2023 11:01:33 GMT
[...]

PR #1552, opened 2024-01-15
Trim trailing empty lines from CLI output #1547

hydromelvictor:


jkbrzt:
Thanks, @hydromelvictor, but trimming trailing lines from CLI output is not something we want to introduce. Also, this PR doesn’t achieve that.

PR #1553, opened 2024-01-23
migrate setup.py to setup.cfg

deronnax:
[Setuptools recommends to use declarative setup.cfg over custom imperative setup.py](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#:~:text=This%20approach%20not%20only%20allows%20automation%20scenarios%20but%20also%20reduces%20boilerplate%20code%20in%20some%20cases.) (especially that this one grown some significant body hair)
On the other hand, using `data_files` (that is used here for man pages) is deprecated and has no direct equivalent in declarative package definition, be it `setup.cfg` or `pyproject.toml`, so we might just not be able to replicate the thing.


jkbrzt:
@deronnax thanks for the PR! If we find a way to maintain full compatibility including `data files`, I’ll be happy to merge it.

jkbrzt:
@deronnax thanks! Do we actually need to keep `setup.py` with the `setup()` call around?

deronnax:
oops sorry. Actually that's tox that's depending colorama.

PR #1557, opened 2024-02-07
docs: replace redirection with tee command with sudo for file creation

vostok92:
This pull request addresses an issue where using `sudo`-prefixed `echo` does not grant elevated permissions for the subsequent redirection operation, resulting in a failure to write files under `/etc/apt`.
By replacing the redirection with a `sudo`-prefixed `tee` command, this change ensures that files are written with root privileges, effectively resolving the permission issue. This modification is essential for operations requiring write access to protected directories.


jkbrzt:
Thanks, @vostok92!

Issue #1583, opened 2024-06-20
HTTPS connections failing since requests version 2.32.3


timsutton:
👋 thanks for taking a look :)

## Checklist

- [x] I've searched for similar issues.
- [x] I'm using the latest version of HTTPie.

---

## Minimal reproduction code and steps

1. (create new virtualenv)
2. `pip install httpie`
3. `https https://raw.githubusercontent.com/Homebrew/homebrew-core/HEAD/Formula/h/httpie.rb`

## Current result

```
https: error: SSLError: HTTPSConnectionPool(host='raw.githubusercontent.com', port=443): Max retries exceeded with url: /Homebrew/homebrew-core/HEAD/Formula/h/httpie.rb (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)'))) while doing a GET request to URL: https://raw.githubusercontent.com/Homebrew/homebrew-core/HEAD/Formula/h/httpie.rb
```
## Expected result

Request would succeed.

## Debug output

Please re-run the command with `--debug`, then copy the entire command & output and paste both below:

```bash
$ https --debug https://raw.githubusercontent.com/Homebrew/homebrew-core/HEAD/Formula/h/httpie.rb
https --debug https://raw.githubusercontent.com/Homebrew/homebrew-core/HEAD/Formula/h/httpie.rb
HTTPie 3.2.2
Requests 2.32.3
Pygments 2.18.0
Python 3.12.3 (main, Jun  8 2024, 18:03:49) [Clang 15.0.0 (clang-1500.3.9.4)]
/private/tmp/.venv-pyenv-py3/bin/python3
Darwin 23.5.0

<Environment {'apply_warnings_filter': <function Environment.apply_warnings_filter at 0x10219a160>,
 'args': Namespace(),
 'as_silent': <function Environment.as_silent at 0x10219a020>,
 'colors': 256,
 'config': {'default_options': []},
 'config_dir': PosixPath('/Users/tsutton/.config/httpie'),
 'devnull': <property object at 0x10217f560>,
 'is_windows': False,
 'log_error': <function Environment.log_error at 0x10219a0c0>,
 'program_name': 'https',
 'quiet': 0,
 'rich_console': <functools.cached_property object at 0x102114ec0>,
 'rich_error_console': <functools.cached_property object at 0x10109d580>,
 'show_displays': True,
[...]

PR #1596, opened 2024-09-04
Explicitly load default certificates when creating SSL context (#1583)

AdamWill:
Requests prior to 2.32.3 always loaded the default (system-wide) set of trusted certificates into custom SSL contexts. 2.32.3 no longer does. This has broken a lot of users, but the fix is moving slowly upstream due to security considerations - see https://github.com/psf/requests/issues/6730 and https://github.com/psf/requests/pull/6731 .

As suggested at https://github.com/psf/requests/pull/6710#issuecomment-2137802782 this can be worked around by explicitly loading the default certificates into the context. We check the method exists before calling it just to be safe, it was added in Python 3.4.

Also, drop the dependency pin as it's no longer needed with this workaround.

### DIFF
--- a/httpie/ssl_.py
+++ b/httpie/ssl_.py
@@ -48,6 +48,13 @@ class HTTPieHTTPSAdapter(HTTPAdapter):
             ssl_version=ssl_version,
             ciphers=ciphers,
         )
+        # workaround for a bug in requests 2.32.3, see:
+        # https://github.com/httpie/cli/issues/1583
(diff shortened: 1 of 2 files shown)


AdamWill:
tweaked to only do anything if there are no certs loaded into the context already, just in case there's some way you could cause custom certs to be used via an environment variable or something.

hinricht:
Please merge soon !

IceCodeNew:
I just confronted this problem, really appreciate the fix.  
I am wondering if somebody could help merge this PR.

PR #1598, opened 2024-09-06
Added an alias for HTTPS in fish completions

cloudmaker97:
Ensure that fish completion is working with the `https` command. 
It just wraps the identical completions from `http` to `https`.

[Documentation for fish-functions, see --wraps](https://fishshell.com/docs/current/cmds/function.html#description)
> Inherit completions from the given WRAPPED_COMMAND. See the documentation for complete for more information.


sersorrel:
this unfortunately breaks `https` for fish users. I defined `https` as `http --default-scheme https $argv`, but now the default-scheme argument gets lost, because my function is overwritten by the one defined here. (and httpie is invoked as `http`, so it can't tell that it should do anything different, as [it's documented to](https://httpie.io/docs/cli/other-default-schemes).)

Issue #1597, opened 2024-09-05
Update httpie to use requests lib version >= 2.32.1


ndate46:
Current Issue: 
The current httpie version 3.2.3 has a dependency on requests library and it requires requests version lib to be >=2.22.0 and<=2.31.0

Enhancement Request: 
httpie needs to be updated to have dependency on requests lib version which is >= 2.32.1 

Problem it solves:
requests lib version 2.31.0 is an old version and has few vulnerabilities. Downgrading httpie to 3.2.2 doesn't help either as it has some vulnerabilities as well and its recommended to use 3.2.3

Issue #1581, opened 2024-05-30
Unable to access HTTPS URLs on Windows securely


jaraco:
## Checklist

- [x] I've searched for similar issues.
- [x] I'm using the latest version of HTTPie.

---

## Minimal reproduction code and steps

1. Install httpie using `pipx install httpie`
2. `https google.com`

## Current result

```

https: error: SSLError: HTTPSConnectionPool(host='google.com', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)'))) while doing a GET request to URL: https://google.com/
```

## Expected result

The resource should be loaded (ideally securely). It should be easy to use. Error messages should be informative and give the user direction on how to remedy the problem. Install docs should provide guidance on how to remedy the situation.

---

## Debug output

Please re-run the command with `--debug`, then copy the entire command & output and paste both below:

```bash
 ~ # https --debug google.com
HTTPie 3.2.2
Requests 2.32.3
Pygments 2.18.0
Python 3.12.2 (tags/v3.12.2:6abddd9, Feb  6 2024, 21:26:36) [MSC v.1937 64 bit (AMD64)]
C:\Users\jaraco\AppData\Local\pipx\pipx\venvs\httpie\Scripts\python.exe
Windows 11

<Environment {'apply_warnings_filter': <function Environment.apply_warnings_filter at 0x000002248A99AA20>,
 'args': Namespace(),
 'as_silent': <function Environment.as_silent at 0x000002248A99A8E0>,
 'colors': 256,
 'config': {'default_options': []},
 'config_dir': WindowsPath('C:/Users/jaraco/AppData/Roaming/httpie'),
 'devnull': <property object at 0x000002248A946020>,
 'is_windows': True,
 'log_error': <function Environment.log_error at 0x000002248A99A980>,
 'program_name': 'https',
 'quiet': 0,
 'rich_console': <functools.cached_property object at 0x000002248A974260>,
 'rich_error_console': <functools.cached_property object at 0x000002248A975400>,
 'show_displays': True,
 'stderr': <colorama.ansitowin32.StreamWrapper object at 0x000002248A024D10>,
 'stderr_isatty': True,
 'stdin': <_io.TextIOWrapper name='<stdin>' mode='r' encoding='utf-8'>,
 'stdin_encoding': 'utf-8',
[...]

PR #1601, opened 2024-09-16
chore (setup.cfg): update request to  2.32.1

aarcex3:
# Related Issue
#1597

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,7 +50,7 @@ install_requires =
     pip
     charset_normalizer>=2.0.0
     defusedxml>=0.6.0
-    requests[socks] >= 2.32.1
+    requests[socks] >=2.22.0, <=2.32.3
     Pygments>=2.5.2
     requests-toolbelt>=0.9.1
     multidict>=4.7.0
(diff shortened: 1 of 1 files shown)


jkbrzt:
Newer versions of Requests have an issue with the way we pass our [custom SSL context](https://github.com/httpie/cli/blob/011402152c696dafda045c6d7aed7c50a155d14b/httpie/ssl_.py#L54). That needs to be fixed before we can upgrade.

See https://github.com/httpie/cli/issues/1583 https://github.com/httpie/cli/issues/1581

Example:

```sh
$ https google.com
```
`https: error: SSLError: HTTPSConnectionPool(host='google.com', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)'))) while doing a GET request to URL: https://google.com/`

jkbrzt:
Fixed in v3.2.4 via #1596

Issue #758, opened 2019-03-01
Define http request in a .http file


tonsV2:
First of all, thanks for a great http client. I've recommended it to quite a few coworkers by now and a lot of appreciation has been returned!

Intellij has a build in http client which supports something called .http files.

https://blog.jetbrains.com/phpstorm/2017/09/editor-based-rest-client/

To sum up the above link you can define http request in a file and have them executed.

Being able to define a file with various http requests and have that file executed by httpie would be so awesome! This could be used for stress testing, integration testing or executed in a CI pipeline for various reasons and probably several other use cases.

Any comments?

GotoCode:
I think something similar was suggested by @dsego in [this comment](https://github.com/jakubroztocil/httpie/issues/537#issuecomment-264718507) on #537, so there might be quite a bit of support for such a feature.

jakubroztocil:
@GotoCode it’s coming—stay tuned!

lisitsky:
Awesome feature for autotests! Can I help somehow with it?

danslapman:
+ https://github.com/Huachao/vscode-restclient - visual studio code have a similiar (but not 100% compatible with IDEA) *.http files support

stonecharioteer:
Is this feature still on the roadmap?

jakubroztocil:
@stonecharioteer Yes, it is! No timeline yet, though.

ferenczy:
I have proposed a solution for such a feature in [this feature request](https://github.com/httpie/httpie/issues/1047), maybe you'll find some inspiration there.

tonsV2:
@jakubroztocil Are there any updates on this matter?

I've been needing this for quite a few years now and there might be people here, myself included, which would be willing to work.

vtronko:
I guess, it's off the roadmap since Desktop app has more priority?

devsdmf:
Do we have any expectations to prioritize this feature in the following weeks/months? I can help with the code if no one already started to work on that. Can I get a green light?

adriens:
:heavy_plus_sign: :one: for this feature request :heavy_heart_exclamation:

afonsocarlos:
I'm also really interested in that one

vzlobinsky:
Do we have any news regarding this feature? I assume many people need it and will love it :)

adriens:
> Do we have any news regarding this feature? I assume many people need it and will love it :)

Yes, please make this happen :pray:

hbjydev:
[...]

PR #1619, opened 2025-03-07
Added support for defining http requests in a .http file

AlessandroColi:
As asked by issue #758 we added basic support for defining requests in a .HTTP file by using the --http-file flag followed by the file name instead of the URL

### DIFF
--- a/httpie/cli/argparser.py
+++ b/httpie/cli/argparser.py
@@ -203,6 +203,8 @@ class HTTPieArgumentParser(BaseHTTPieArgumentParser):
         }
 
     def _process_url(self):
+        if self.args.http_file:
+            return
         if self.args.url.startswith('://'):
             # Paste URL & add space shortcut: `http ://pie.dev` → `http://pie.dev`
             self.args.url = self.args.url[3:]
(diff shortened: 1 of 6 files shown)


bittner:
Tests are failing. Unfortunately, the logs are gone now as well. Can you make the test suite pass?

httpie/cli/argparser.py
```diff
@@ -203,6 +203,10 @@ def _process_request_type(self):
         }
 
     def _process_url(self):
+        if self.args.http_file:
+            # do not add default scheme
+            # treat URL as a filename if --http-file is specified
+            return
```
bittner, line 209: I wonder if `--file` wouldn't be nicer to read.

```console
http --http-file example.http
```

vs.

```console
http --file example.http
```

## Read from stdin

An elegant alternative – the nicest way, in my opinion – would be to read from standard input, e.g.

Redirect a file:

```console
$ http < example.http
```

Parse some text verbatim:

```console
$ echo "
    GET http://localhost:8080/list
    Accept: application/json
" | http
```

Interpret input typed or pasted in the console:

```console
$ http
GET http://localhost:8080/list
Accept: application/json
^D
```

PR #1654, opened 2025-08-11
download: fix progress with gzip + regression test"

ali90h:
Minimal fix + regression test.

When Content-Encoding is present, we no longer trust Content-Length for progress.
We track received bytes instead, which prevents false 'Incomplete download' and >100% overshoot.

- Focused change in httpie/downloads.py (cherry-picked from #1653)
- Adds a dedicated regression test: gzip + Content-Length case
- No unrelated CI/docs/style changes"


ali90h:
Hi @isidentical could you please review when you have a moment?
This PR is very small: a focused fix in httpie/downloads.py plus a regression test for the gzip + Content-Length case.
CI failures look unrelated (3.7 runner / Big5 encoding heuristics). I’m happy to open a separate CI PR if needed. Thanks!

PR #1691, opened 2026-01-24
Normalize Content-Length header to int

joshmuepu:
### Summary
This change normalizes `Content-Length` header values before upload preparation so that values provided as strings or bytes are converted to an integer early and consistently.

This prevents incorrect buffering and streaming behavior when `Content-Length` is present but not an `int`.

### Details
- Coerces `Content-Length` values to `int` when possible
- Raises `ValueError` for invalid values instead of silently continuing
- Keeps behavior unchanged for already-valid integer values

### Tests
- Existing upload-related tests
- Manual verification of request preparation paths

### Notes
While running the full test suite locally, I observed failures in `tests/test_encoding.py` related to Big5 charset detection that appear to be environment-specific and unrelated to this change.


joshmuepu:
CI failures appear unrelated to this change.

I reproduced the Big5 encoding failures locally and they also fail across Linux, macOS, and Windows in CI. The changes in this PR are limited to Content-Length normalization in uploads.py and do not affect encoding or CLI argument parsing paths.

Happy to rebase or adjust if you’d prefer this to land after CI issues are addressed upstream.

PR #1698, opened 2026-02-18
Improve error message for malformed JSON in request body

ktalatiuf:


ktalatiuf:
This PR improves the error message shown when malformed JSON is provided in the request body by making it more user friendly. Modified the load_json function.

PR #1817, opened 2026-04-14
Fix terminal escape sequence injection via HTTP response data

### LINKED ISSUE #1812: Terminal escape sequence injection via malicious HTTP response data
# Terminal escape sequence injection via HTTP response data

## Summary

HTTPie writes HTTP response headers and body content to the terminal without stripping or escaping terminal control sequences. A malicious HTTP server can embed ANSI escape codes or other terminal control characters in response headers or body text, potentially manipulating the user's terminal display, changing the terminal title, injecting clipboard content (on supported terminals), or in worst-case scenarios with vulnerable terminal emulators, executing commands.

## Details

When HTTPie renders a response to a TTY, the data flows through `output/streams.py` and `output/writer.py` without any sanitization of terminal control characters. While Pygments-based colorization adds its own ANSI escape sequences for syntax highlighting, it does not filter out escape sequences that are already present in the response data.

The relevant code path is:

1. Response received from server with embedded escape sequences
2. `BaseStream.__iter__()` yields header and body chunks (`streams.py`, line 63)
3. `write_stream()` writes chunks directly to the output buffer (`writer.py`, line 73)
4. Terminal interprets all embedded escape sequences

Affected locations:
- `httpie/output/streams.py` - `BaseStream.__iter__()`, `EncodedStream`, `PrettyStream`
- `httpie/output/writer.py` - `write_stream()`, `write_stream_with_colors_win()`

## Impact

**Terminal display manipulation**: An attacker controlling a server can overwrite previous terminal output, hide the real response, or display misleading content. For example, a response could clear the screen and display fake content that looks like a different response.

**Terminal title/clipboard injection**: On terminals that support OSC escape sequences:
- `\x1b]2;text\x07` changes the terminal window title
- `\x1b]52;c;base64\x07` sets clipboard content on some terminals (iTerm2, some xterm configs)

**Terminal emulator vulnerabilities**: Some terminal emulators have had historical vulnerabilities where specific escape sequences can trigger command execution or file writes (e.g., CVE-2003-0063 for xterm window title reporting, various iTerm2 and Konsole issues).

This is particularly relevant for HTTPie because it's commonly used in security research and API testing, contexts where users frequently interact with untrusted servers.

## Reproduction

Start a malicious server:

```python
import http.server

class Handler(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        # Inject terminal title change via header
        self.send_header('X-Custom', '\x1b]2;INJECTED TITLE\x07normal value')
        self.end_headers()
        # Body: clear screen + fake content
        body = '\x1b[2J\x1b[HHTTP/1.1 200 OK\r\nContent-Type: text/plain\r\n\r\nEverything is fine.'
        self.wfile.write(body.encode())

http.server.HTTPServer(('localhost', 9999), Handler).serve_forever()
```

Then:
```bash
http localhost:9999
```

The terminal title will change, and the real response content will be replaced with fake output.

## Suggested Fix

Filter terminal control characters when outputting to a TTY. A conservative approach:

```python
import re

# Match ANSI escape sequences and other control characters
CONTROL_CHAR_RE = re.compile(
    r'[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]'  # C0 control chars (except \t, \n, \r)
    r'|\x1b(?:\[[^a-zA-Z]*[a-zA-Z]'          # CSI sequences
    r'|\][^\x07]*\x07'                         # OSC sequences
    r'|\(.'                                     # charset selection
    r'|.)'                                      # other escape sequences
)

def sanitize_for_terminal(text: str) -> str:
    return CONTROL_CHAR_RE.sub('', text)
```

Apply this sanitization in the stream output path when `env.stdout_isatty` is True, but not when output is redirected to a file or pipe (to preserve raw data for scripting).

This could be implemented as a flag (e.g., `--raw-terminal` to disable sanitization) for users who explicitly want raw output.

eddieran:
## Summary

Fixes #1812 — Terminal escape sequence injection via HTTP response data.

A malicious HTTP server can embed ANSI escape sequences in response headers or body to manipulate the user's terminal when HTTPie displays the output. This can be used to:

- Set the terminal window title (OSC 0)
- Inject clipboard content (OSC 52)
- Move the cursor to overwrite previously displayed content (CSI H)
- Clear the screen (CSI J)
- Reset the terminal (ESC c)

### Fix

Added `httpie/output/sanitize.py` with a `sanitize_output()` function that strips dangerous terminal control sequences from output bytes while preserving:

- Safe whitespace: `\t`, `\n`, `\r`
- SGR color codes (`ESC [ ... m`) used by HTTPie's own syntax highlighting

The sanitization is applied in `write_stream()` and `write_stream_with_colors_win()` only when output is to a TTY (`stdout_isatty=True`). Piped/redirected output is left untouched so raw data is preserved for scripts.

### What is stripped

| Sequence type | Example | Purpose |
|---|---|---|
| OSC sequences | `ESC ] 0;title BEL` | Window title, clipboard manipulation |
| Non-SGR CSI sequences | `ESC [ 2 J`, `ESC [ 1;1 H` | Screen clear, cursor movement |
| Single-char escapes | `ESC c` | Terminal reset |
| Dangerous C0 chars | `BEL`, `BS`, `VT`, `FF`, `DEL` | Bell, backspace, etc. |

### Files changed

- **`httpie/output/sanitize.py`** — New module with `sanitize_output()` function
- **`httpie/output/writer.py`** — Apply sanitization when writing to TTY
- **`tests/test_terminal_sanitize.py`** — 19 tests (14 unit + 5 integration)

## Test plan

- [x] All 19 new tests pass
- [x] All 137 existing output/stream tests pass (0 regressions)
- [ ] Manual test: `printf 'HTTP/1.1 200\r\nX-Evil: \x1b]0;pwned\x07\r\n\r\nbody' | nc -l 8888` then `http localhost:8888` — terminal title should not change

### DIFF
--- a/httpie/output/sanitize.py
+++ b/httpie/output/sanitize.py
@@ -1,0 +1,68 @@
+"""
+Sanitize terminal control sequences from HTTP response output.
+
+When outputting to a TTY, malicious HTTP responses can inject terminal
+escape sequences (ANSI CSI, OSC, etc.) to manipulate the terminal display,
+change the window title, inject clipboard content, or worse.
+
+This module provides sanitization that strips dangerous control characters
+while preserving safe whitespace (\t, \n, \r) and legitimate color codes
+added by HTTPie's own formatting/syntax highlighting.
+
+See: https://github.com/httpie/cli/issues/1812
+"""
+import re
+
+# CSI (Control Sequence Introducer) sequences: ESC [ ... <final byte>
+# These control cursor movement, display attributes, etc.
+# We preserve color sequences (SGR: ESC [ <params> m) since HTTPie adds those.
+_CSI_RE = re.compile(
+    rb'\x1b\['           # ESC [
+    rb'[0-9;]*'          # parameter bytes
+    rb'[^0-9;m]'         # final byte that is NOT 'm' (SGR)
+)
+
+# OSC (Operating System Command) sequences: ESC ] ... (ST | BEL)
+# These can set window title, manipulate clipboard, etc.
+_OSC_RE = re.compile(
+    rb'\x1b\]'           # ESC ]
+    rb'.*?'              # payload
+    rb'(?:\x1b\\|\x07)'  # string terminator (ESC \ or BEL)
+)
+
+# Other escape sequences: ESC followed by a single character
+# (e.g., ESC 7, ESC 8, ESC c for terminal reset)
+# Exclude ESC [ (CSI, handled above) and ESC ] (OSC, handled above)
+_OTHER_ESC_RE = re.compile(
+    rb'\x1b'             # ESC
+    rb'[^[\]]'           # any char except [ and ]
+)
+
+# C0 control characters that are dangerous.
+# We keep: \t (0x09), \n (0x0a), \r (0x0d), \x1b (handled by regex above)
+# We strip: all other C0 chars (0x00-0x08, 0x0b-0x0c, 0x0e-0x1a, 0x1c-0x1f)
+# Plus DEL (0x7f)
+_C0_DANGEROUS = re.compile(
+    rb'[\x00-\x08\x0b\x0c\x0e-\x1a\x1c-\x1f\x7f]'
+)
+
+
+def sanitize_output(data: bytes) -> bytes:
+    """Remove dangerous terminal control sequences from output bytes.
+
+    Preserves:
+    - Tab (\\t), newline (\\n), carriage return (\\r)
+    - SGR color sequences (ESC [ ... m) used by HTTPie's syntax highlighting
+
+    Removes:
+    - CSI sequences other than SGR (cursor movement, scrolling, etc.)
+    - OSC sequences (window title, clipboard manipulation, etc.)
+    - Other single-character escape sequences (terminal reset, etc.)
+    - Dangerous C0 control characters and DEL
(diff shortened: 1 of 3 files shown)


eddieran:
Heads-up on the red CI: the two failing jobs are `test_encoding.py::test_terminal_output_{response,request}_charset_detection[big5-...]`, and they're **not** caused by this change.

I checked out the PR's base commit (`5b604c3`, none of my changes) and ran the same test in the same env — it fails identically there:

```
assert '卷首卷首…' in 'HTTP/1.1 200 OK…뻥솤뻥솤…'
```

It's `charset_normalizer` (3.4.7 here) detecting the big5 sample as euc-kr/cp949, so the round-trip produces Korean glyphs instead of the expected Chinese. The test data hasn't changed on `master` since 2024, so any PR built on a runner with a recent `charset_normalizer` will trip it.

This PR's own additions (`tests/test_terminal_sanitize.py`) pass. The sanitizer only touches the byte stream on TTY output and never sees ESC/C0 bytes inside big5 sequences, so it doesn't affect charset detection. Happy to pin `charset_normalizer` in a separate PR if that'd help unblock the suite.

PR #1832, opened 2026-04-21
feat: add default_timeout support to config

Marias03:
## What does this PR do?
Adds support for `default_timeout` in HTTPie's config file.
Users can now set a global timeout that applies to all requests automatically.

## How to use it?
Add to your `config.json`:
```json
{
    "default_timeout": 30
}
```

This timeout will apply to all requests unless overridden with `--timeout`.

## Changes
- `httpie/config.py`: Added `default_timeout` to `Config.DEFAULTS` and a new property
- `httpie/client.py`: Applied `default_timeout` from config in `make_send_kwargs`
- `tests/test_config.py`: Added two tests for the new feature

### DIFF
--- a/httpie/client.py
+++ b/httpie/client.py
@@ -63,7 +63,7 @@ def collect_messages(
         base_headers=httpie_session_headers,
         request_body_read_callback=request_body_read_callback
     )
-    send_kwargs = make_send_kwargs(args)
+    send_kwargs = make_send_kwargs(args, env)
     send_kwargs_mergeable_from_env = make_send_kwargs_mergeable_from_env(args)
     requests_session = build_requests_session(
         ssl_version=args.ssl_version,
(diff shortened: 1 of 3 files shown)


httpie/client.py
```diff
@@ -343,6 +346,9 @@ def make_request_kwargs(
     headers = make_default_headers(args)
     if base_headers:
         headers.update(base_headers)
+    # Apply default headers from config (can be overridden by user)
+    if env.config.default_headers:
```
mariamanbar, line 350: suggestion: safely guard the `env` reference to prevent NoneType attribute errors

While `env` is safely guarded up in `make_send_kwargs`, it is missing a guard here in `make_request_kwargs` on line 350:

```python
if env and env.config.default_headers:

File: tests/test_config.py [hunk 2]
```diff
@@ -102,3 +102,36 @@ def test_custom_config_dir(monkeypatch: MonkeyPatch, tmp_path: Path):
...
+    env.config.save()
+    assert env.config.default_timeout == 30
+
+
+def test_default_timeout_overridable(httpbin):
+    env = MockEnvironment()
+    env.config['default_timeout'] = 30
+    env.config.save()
+    r = http('--timeout=60', httpbin + '/get', env=env)
+    assert HTTP_OK in r
```
mariamanbar, line 137: suggestion: add a specific assertion to verify the timeout override value

While checking for `HTTP_OK` confirms the request succeeded, the test doesn't explicitly verify that the `--timeout=60` argument successfully overrode the `default_timeout` setting of `30` during execution. 

Adding an assertion for checking would make this test block significantly stronger.

PR #1891, opened 2026-07-05
argparser fixes and add tests.

batuhansenol:
I fixed FIXME on argparse.py and added some tests.


batuhansenol:
The GitHub Actions workflow needs to be updated.

PR #1893, opened 2026-07-07
Preserve non-canonical numeric nested JSON keys

upuddu:
HTTPie's nested-JSON syntax decides whether a key segment is an array index or a text key by calling `int()` on it. `int()` is lenient - it also accepts `007`, `1_000`, `+5`, and whitespace-padded values - so numeric-looking object keys were silently rewritten. `http pie.dev/post 007=1` sent `{"7": "1"}` instead of `{"007": "1"}`, quietly dropping the leading zeros; `1_000` became `1000` and `+5` became `5`.

This only treats a segment as a number when it's written in canonical integer form, keeping everything else as verbatim text. Canonical values including negative indexes like `foo[-10]` are unaffected. Added coverage to `test_nested_json_syntax` and a changelog entry.


upuddu:
The failing checks here are pre-existing on `master` (5b604c3) and unrelated to this PR, which only touches `httpie/cli/nested_json/parse.py` and its test:

- `test_encoding.py::test_terminal_output_*_charset_detection[big5-…]` fails because current `charset_normalizer` (3.4.x) detects the big5 sample as EUC-KR. It also fails on unmodified `master`.
- `test_auth.py::test_digest_auth[https-…]` fails intermittently with `Context has already been used to create a Connection` (urllib3/pytest-httpbin SSLContext reuse).
- The `3.7` matrix legs fail at setup because GitHub runners no longer provide Python 3.7; the `macos-13` legs time out at the runner level.

All added tests pass locally (`pytest tests/test_json.py` → 295 passed). Repo-level fixes for the above (dependency pins / dropping 3.7 / SSLContext test fix) belong in separate infra PRs, not here. Happy to rebase once master's CI baseline is repaired.

PR #1899, opened 2026-07-15
Install fish shell completions via pip

### LINKED ISSUE #912: fish shell completions
It would be amazing if httpie could default-install completions for the awesome fish shell.

Reference: https://fishshell.com/docs/current/#where-to-put-completions

Example of other great tools that does this:

```sh
ll  /usr/share/fish/vendor_completions.d
.rw-r--r-- 4.4k root root 25 Apr 13:48 bat.fish
.rw-r--r--  51k root root 11 Mar  2:24 docker.fish
.rw-r--r--  12k root root  9 May 17:38 rg.fish
```

yoshibase:
## Summary

Fixes #912.

Install HTTPie's existing Fish completions through Python wheels and source distributions, so Fish discovers completions for the `http` command without downstream package-specific copying.

## Changes

- Add `share/fish/vendor_completions.d` to `setup.cfg` data files
- Install the canonical `extras/httpie-completion.fish` unchanged
- Add a command-named `http.fish` loader that sources the canonical definitions, avoiding a duplicated completion script
- Include both files in the source distribution manifest

## Test plan

- [x] `python -m build --sdist --wheel`
- [x] Wheel contains `share/fish/vendor_completions.d/http.fish`
- [x] Wheel contains `share/fish/vendor_completions.d/httpie-completion.fish`
- [x] Source distribution contains both source files

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,3 +6,7 @@ include docs/README.md
 
 # <https://github.com/httpie/cli/issues/182>
 recursive-include tests/ *
+
+# Shell completions installed through setup.cfg data_files
+include extras/httpie-completion.fish
+include extras/fish/http.fish
--- a/extras/fish/http.fish
+++ b/extras/fish/http.fish
@@ -1,0 +1,2 @@
+# Fish loads completions by command name; keep the canonical definitions in one file.
+source (dirname (status --current-filename))/httpie-completion.fish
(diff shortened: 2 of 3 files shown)


yoshibase:
CI note after the clean rebuild: the current PR is only the 3 packaging/completion files, and the wheel + sdist build succeeds with both Fish files in `share/fish/vendor_completions.d`.

The remaining red jobs reproduce on upstream `master` (for example [run 28435490851](https://github.com/httpie/cli/actions/runs/28435490851)): Big5 detection assertions, Python 3.7 unavailable on Ubuntu 24.04, argparse formatting drift, and an unrelated urllib3 digest-auth error. None touches this PR's files.

PR #1900, opened 2026-07-15
Fix Windows guard in man page availability check

### LINKED ISSUE #1898: Bug: Windows guard in is_available() compares os.system (a function) to 'nt' — always False
## Description

```python
        return False
    ...
```

## Why it is wrong

## Impact

```python
import os
```

## Suggested fix

```python
import sys
...
        return False
    ...
```

## Affected version

yoshibase:
## Summary

Fixes #1898.

`is_available()` compared the `os.system` function object to `"nt"`, so the intended Windows short-circuit could never run. Use `sys.platform == "win32"` and cover every return path.

## Test plan

- [x] `pytest tests/test_man_pages.py` — 5 passed
- [x] `flake8 httpie/output/ui/man_pages.py tests/test_man_pages.py`
- [x] Windows guard, `HTTPIE_NO_MAN_PAGES`, successful/failed `man`, and execution-error paths covered

### DIFF
--- a/httpie/output/ui/man_pages.py
+++ b/httpie/output/ui/man_pages.py
@@ -2,6 +2,7 @@
 
 import subprocess
 import os
+import sys
 from httpie.context import Environment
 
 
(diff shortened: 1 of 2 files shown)


yoshibase:
CI note after the clean rebuild: `tests/test_man_pages.py` passes all 5 cases and flake8 is clean. The current PR is only the man-page guard plus its focused test file.

The coverage failure is the same pre-existing Big5 detection failure in `tests/test_encoding.py` seen on upstream `master` (for example [run 28435490851](https://github.com/httpie/cli/actions/runs/28435490851)); the new man-page tests report 100% coverage.

PR #1903, opened 2026-07-21
Fix Windows man pages guard (os.system to sys.platform)

### LINKED ISSUE #1898: Bug: Windows guard in is_available() compares os.system (a function) to 'nt' — always False

vtino17:
Fixes #1898

## Problem

`os.system` is a function object, so comparing it with `"nt"` is always false. HTTPie can therefore attempt to invoke `man` on Windows.

## Change

- use `sys.platform == "win32"` for the platform guard
- return before spawning a subprocess on Windows
- add a regression test proving `subprocess.run` is not called

## Validation

- focused Windows guard regression: pass
- code-style check: pass

### DIFF
--- a/httpie/output/ui/man_pages.py
+++ b/httpie/output/ui/man_pages.py
@@ -2,6 +2,7 @@
 
 import subprocess
 import os
+import sys
 from httpie.context import Environment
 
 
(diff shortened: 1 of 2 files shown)


vtino17:
I pushed a focused regression test in commit 3a545d2. It verifies that Windows returns false before attempting to invoke man, while the implementation now uses sys.platform == "win32". The isolated Python regression check passes; CI is the cross-platform authority.

Issue #1623, opened 2025-03-18
Username/password in the URL not decoded when used for basic authentication


vadz:
## Checklist

- [x] I've searched for similar issues.
- [x] I'm using the latest version of HTTPie.

---

## Minimal reproduction code and steps

Run `http https://u%40d:1%3d2%3f@httpbin.org/basic-auth/u%40d/1%3d2%3f`.

## Current result

401 status is returned because HTTPie tries to authenticate using `u%40d` username instead of the expected `u@d` (and wrong password, too, for the same reason).

## Expected result

Same as from running `http -a 'u@d:1=2?' https://httpbin.org/basic-auth/u%40d/1%3d2%3f`, i.e. 200 status response.

---

## Debug output

Please re-run the command with `--debug`, then copy the entire command & output and paste both below:

```bash
$ http --debug https://u%40d:1%3d2%3f@httpbin.org/basic-auth/u%40d/1%3d2%3f
HTTPie 3.2.4
Requests 2.32.3
Pygments 2.19.1
Python 3.11.2 (main, Nov 30 2024, 21:22:50) [GCC 12.2.0]
/home/zeitlin/python/3.11/bin/python3
Linux 6.7.9-amd64

<Environment {'apply_warnings_filter': <function Environment.apply_warnings_filter at 0x7ff3af8a0720>,
 'args': Namespace(),
 'as_silent': <function Environment.as_silent at 0x7ff3af8a05e0>,
 'colors': 256,
 'config': {'default_options': []},
 'config_dir': PosixPath('/home/zeitlin/.config/httpie'),
 'devnull': <property object at 0x7ff3af87d4e0>,
 'is_windows': False,
 'log_error': <function Environment.log_error at 0x7ff3af8a0680>,
 'program_name': 'http',
 'quiet': 0,
 'rich_console': <functools.cached_property object at 0x7ff3af8916d0>,
 'rich_error_console': <functools.cached_property object at 0x7ff3af891750>,
 'show_displays': True,
 'stderr': <_io.TextIOWrapper name='<stderr>' mode='w' encoding='utf-8'>,
 'stderr_isatty': False,
 'stdin': <_io.TextIOWrapper name='<stdin>' mode='r' encoding='utf-8'>,
 'stdin_encoding': 'utf-8',
 'stdin_isatty': True,
 'stdout': <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>,
 'stdout_encoding': 'utf-8',
[...]

PR #1936, opened 2026-08-13
Fixed URL-encoded characters in basic auth credentials (#1623)

abhishekjoshi13:
When providing credentials in the URL that contain URL-encoded characters (like %40 for @), the urlsplit output was not being decoded before being passed to AuthCredentials. This resulted in the raw encoded string being Base64 encoded, causing server rejection. I injected urllib.parse.unquote in argparser.py to decode url.username and url.password prior to generating the Basic Auth headers


abhishekjoshi13:
HI team, just a quick heads up. The fix for decoding credentials in the URL is pushed and ready for review. I noticed the CI pipeline is throwing some failures, but after reviewing the server logs, they appear to be unrelated upstream issues:
1. test_cli_ui.py::test_naked_invocation - Failing on python 3.12 due to the argparse formatting changes(quotes being added around choices)
2. test_encoding.py - The big5 charset tests appear to be failing across environments.
Let me know if you would like me to write a dedicated unit test for this URL decoding fix, or if there is anything else you need from me to get this merged.

PR #1944, opened 2026-08-29
Turn a bare AssertionError into a real error message

### LINKED ISSUE #1614: AssertionError when trying to POST with bearer auth
## Checklist

- [X] I've searched for similar issues.
- [X] I'm using the latest version of HTTPie.

---

```
/Users/Khayet https POST --auth-type bearer --auth $TOKEN "https://example.org"
```

## Current result
```
  File "/opt/homebrew/bin/http", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/__main__.py", line 9, in main
    exit_status = main()
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/core.py", line 162, in main
    return raw_main(
        parser=parser,
    ...<2 lines>...
        env=env
    )
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/core.py", line 77, in raw_main
    parsed_args = parser.parse_args(
        args=args,
        env=env,
    )
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/cli/argparser.py", line 176, in parse_args
    self._guess_method()
    ~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/cli/argparser.py", line 416, in _guess_method
    assert not self.args.request_items
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

## Expected result

Probably an error message telling me what is wrong about the invocation and how to do it correctly. I'm new to httpie and guess that I did something wrong but I don't know what. `https --help` is telling me that the method parameter comes directly after the `https` command but maybe the problem is that the url is in the wrong place and httpie gets confused due to the auth parameters?

## Debug output

Please re-run the command with `--debug`, then copy the entire command & output and paste both below:

```bash
/Users/Khayet https POST --auth-type bearer --auth $TOKEN "https://example.org" --debug
HTTPie 3.2.4
Requests 2.32.3
Pygments 2.18.0
Python 3.13.1 (main, Dec  3 2024, 17:59:52) [Clang 16.0.0 (clang-1600.0.26.4)]
/opt/homebrew/Cellar/httpie/3.2.4/libexec/bin/python
Darwin 24.3.0

<Environment {'apply_warnings_filter': <function Environment.apply_warnings_filter at 0x106628220>,
 'args': Namespace(),
 'as_silent': <function Environment.as_silent at 0x1066280e0>,
 'colors': 256,
 'config_dir': PosixPath('/Users/jakob.herpel/.config/httpie'),
 'devnull': <property object at 0x106609530>,
 'is_windows': False,
 'log_error': <function Environment.log_error at 0x106628180>,
 'program_name': 'http',
 'quiet': 0,
 'rich_console': <functools.cached_property object at 0x105ba1fd0>,
 'rich_error_console': <functools.cached_property object at 0x1066005a0>,
 'show_displays': True,
 'stderr_isatty': True,
 'stdin_encoding': 'utf-8',
 'stdin_isatty': True,
 'stdout_encoding': 'utf-8',
 'stdout_isatty': True}>

 'converters': [],
  File "/opt/homebrew/bin/http", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/__main__.py", line 9, in main
    exit_status = main()
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/core.py", line 162, in main
    return raw_main(
        parser=parser,
    ...<2 lines>...
        env=env
    )
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/core.py", line 77, in raw_main
    parsed_args = parser.parse_args(
        args=args,
        env=env,
    )
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/cli/argparser.py", line 176, in parse_args
    self._guess_method()
    ~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/httpie/3.2.4/libexec/lib/python3.13/site-packages/httpie/cli/argparser.py", line 416, in _guess_method
    assert not self.args.request_items
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

…

afonsojanu:
Fixes #1614

`http POST --auth-type bearer --auth token URL` crashes with a bare `AssertionError` on Python 3.13, instead of running the request or telling the caller what's wrong with the invocation.

The cause is how argparse buckets tokens once an optional argument sits between METHOD/URL and the first REQUEST_ITEM. It splits the positional run into two separate pieces and, on 3.13, fills each piece's slots independently: `POST` and everything up to the first option land in the METHOD/URL slots, and the actual URL after the options gets read back as the first REQUEST_ITEM, leaving METHOD unset. `_guess_method` assumed METHOD being unset always meant `request_items` was still empty and asserted exactly that, so this specific misparse hit the assertion directly.

The assertion is now a real error explaining that METHOD, URL and REQUEST_ITEM have to sit next to each other with nothing else in between, which is what the report itself expected to see instead of a traceback.

Added a regression test with the exact command from the issue, checking it exits with an error and prints the new message rather than crashing. Ran the full suite on Python 3.13 (where this reproduces) plus 3.11: 1018 passed, 2 pre-existing failures unrelated to this change (big5 charset detection in test_encoding.py, reproduces identically on unmodified master). flake8 clean on both changed files.


afonsojanu:
CI caught something the local run on Python 3.13 didn't: the new regression test only reproduces on 3.13+, since that's the version where argparse fills a split positional run the way the report describes. On 3.8 through 3.12 the same command gets rejected earlier, with argparse's own 'unrecognized arguments' error, so the assertion never gets exercised there. Gated the test on Python 3.13+ so it stays meaningful without failing on versions where it can't reproduce the original crash in the first place. Confirmed the full suite is green on both 3.11 (test now skips, 1017 passed) and 3.13 (test passes).

afonsojanu:
Heads up in case it saves time reviewing: the failing checks here (test_cli_ui naked invocation string match, and the two big5 encoding detection tests) aren't touched by this diff at all, they're in test_cli_ui.py and test_encoding.py while this PR only changes argparser.py and test_httpie.py. Looks like an argparse choices-quoting difference between Python patch versions and a locale-dependent chardet result, both pre-existing on this branch's base.

PR #1945, opened 2026-09-04
Report invalid --format-options as a CLI error, not a traceback

seven7763:
## Summary

Fixes #1938.

An invalid `--format-options` value crashed with an unhandled `argparse.ArgumentTypeError` traceback:

```
$ http --offline --format-options=json.indent:x pie.dev/get
argparse.ArgumentTypeError: invalid value 'x' in 'json.indent:x' (expected int got str)
```

`_process_format_options()` calls `parse_format_options()` manually after parsing, so the `ArgumentTypeError` raised for bad input is never converted into a normal CLI error. This catches it there and routes it through `self.error()`, which is the same pattern the parser already uses for other invalid-flag combinations.

```
$ http --offline --format-options=json.indent:x example.org

error:
    invalid value 'x' in 'json.indent:x' (expected int got str)
```

## Changes

- `httpie/cli/argparser.py`: wrap the `parse_format_options()` loop in `_process_format_options()` with `try/except argparse.ArgumentTypeError` → `self.error(str(ex))`
- `tests/test_output.py`: 3 regression tests (bad type / unknown key / bare section), asserting a clean `error:` message, `ExitStatus.ERROR`, and no traceback

## Validation

- [x] Reproduced all three variants from the issue on `main` before the fix (all traceback)
- [x] After the fix: all three produce the standard CLI error and exit 1
- [x] `python -m pytest tests/test_output.py -q` — 123 passed; the one failure (`TestQuietFlag::test_double_quiet_on_error`) also fails on unmodified `main` in this environment and is untouched by this change

## Notes

No message wording changed — the same messages from `parse_format_options()` are surfaced, just without the traceback.


seven7763:
Heads-up on CI: the failures here (encoding/charset tests + coverage) are repo-wide — PR #1944 shows the same 37 failing checks with unrelated changes. Locally `pytest tests/test_output.py` passes (123 passed; the one failure, `TestQuietFlag::test_double_quiet_on_error`, also fails on unmodified main locally).

PR #1951, opened 2026-09-07
docs: fix fork clone command in contributing guide

chenlichao:
## Summary

Fix the fork-cloning command in `CONTRIBUTING.md` to use the current repository name, `cli.git`, and explicitly name the checkout directory `httpie` so the following `cd httpie` command and later paths remain correct.

The guide links to `httpie/cli`, but the clone command still assumes a fork named `httpie`. A newly created fork with GitHub's default name is `<YOU>/cli`, so the documented remote does not exist for that fork.

This is intentionally a one-line, documentation-only change (1 addition, 1 deletion); no runtime behavior changes.

## Verification

- Created a default-named fork of `httpie/cli` and confirmed its parent and name through the GitHub API.
- Tested the documented remote with `git ls-remote`: the old `chenlichao/httpie.git` target failed with “Repository not found”; the corrected `chenlichao/cli.git` target resolved to the same HEAD as upstream `master`. Used Git's per-command SSH-to-HTTPS URL rewrite for these checks because authentication is configured over HTTPS.
- Extracted the updated Getting the code shell commands, substituted the account name, and checked them with `bash -n`.
- Confirmed the explicit clone destination matches the subsequent `cd httpie`.
- `git diff --check` passed; manually reviewed the one-line diff.

The application test suite was not run because only a Git command in the contributor guide changes. No full clone or SSH authentication test was performed.


chenlichao:
Friendly ping for a first review — one-line fix so the contributing guide's clone command matches GitHub's default fork name (`<YOU>/cli.git`) while keeping the following `cd httpie` correct. Docs-only; happy to adjust the wording if you'd prefer a different approach.

